PR #11, opened 2024-10-09
Feature/navigation

mohamedtahaguelzim:
Hi,

I created this scaffold for navigation, as we need it as a base for our app.
For the moment, no tests are implemented, as it is really not intended to be functionnal.
I created "BlankScreens" that should be replaced with your screens.
I also set up the routes in NavigationController.

**Important :** I created empty view models, as I guessed it would make adding your own screens more intuitive. However, they are again not intended to work, and are only present for compilation purposes. Consider replacing these view models by your own, if implemented.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -118,6 +118,7 @@ dependencies {
     implementation(libs.material)
     implementation(libs.androidx.lifecycle.runtime.ktx)
     implementation(platform(libs.compose.bom))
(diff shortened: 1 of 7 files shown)


aminebenji04:
Looks good to me!
Waiting for the complete version but this should be enough for us to implement our functionnalities.
For the tests, I suggest simple Android Tests that check that navigation works well when clicking on buttons that switch screens

Issue #8, opened 2024-10-09
View my profile


elvan03:
View the activities I created with the possibility to edit them. 
View the activities I enrolled in.
view how is it possible to assign activities to people, and how to give distincts privileges to owner.

Frontend + Backend

PR #19, opened 2024-10-10
Feature/profile

elvan03:
Created the user profile's screen. it contains information on the user such as name, profile picture and interests. Solves issue #8


GhaliBerbich:
I linked the Firestore database to the application, specifically the 'profiles' collection with the profile page, so that I can retrieve profile information from the profiles present in the database. I implemented the UserProfileViewModel file as well as the ProfilesRepositoryFirestore file for that. I then used the profileViewModel in the profileScreen to display the informations of the user from the database.

app/src/main/java/com/android/sample/model/ProfilesRepositoryFirestore.kt
aminebenji04, line 1: LGTM!

app/src/main/java/com/android/sample/model/User.kt
aminebenji04, line 1: Why would a photo be of type String? 
We will maybe add other info to the User data class

app/src/main/java/com/android/sample/model/UserProfileViewModel.kt
aminebenji04, line 1: LGTM!

app/src/main/java/com/android/sample/ui/profile/Profile.kt
aminebenji04, line 1: I think that all we have to do is see how it renders on the screen

gradle/wrapper/gradle-wrapper.properties
aminebenji04, line 1: Why would we remove this line?

aminebenji04 (approved):
LGTM ! Just check why you removed the date in gradle/wrapper/gradle-wrapper.properties

PR #20, opened 2024-10-10
Add navigation feature

Yassyass181104:
Added the navigation feature for the CreateActivityScreen and for the EditActivityScreen such that we can navigate back to the overview when we press the arrow, save, edit or delete an activity


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
aminebenji04, line 1: It looks better now with navigation, Looks good to me !

aminebenji04 (approved):

elvan03 (approved):
Agreed, it's much better with navigation. LGTM!

PR #28, opened 2024-10-12
Some changes in main activity

Yassyass181104:
I made some changes in main activity such that we use the correct viewmodel for activities and deleted the one that was used before


aminebenji04:
I think that you didn't pull the latest changes before writing this

PR #29, opened 2024-10-12
feat(profile): add profile creation feature at the Sign-In and SIgn-Up

aminebenji04:
- Added the possibility to create a profile even if you already created your account without creating a profile.

- Added tests with 94% coverage.


GhaliBerbich (approved):
Good modularization and separation of screens for the creation of the profile.

PR #34, opened 2024-10-13
Fix bug in ActivityDetailsScreen

DAFERMOH:
fixed a bug in which the values of the dataclass of the selected were not displayed in activityDetails screen


aminebenji04 (approved):
LGTM!

aminebenji04 (approved):
I think that adding the viewModel as a parameter is a good practice to always be able to know which is the selected Activity

Yassyass181104 (approved):
Very good, the changes look coherent with the rest of the classes concerned. Well done

PR #39, opened 2024-10-14
Feature/profile creation

aminebenji04:
I just changed the logic so that: if we login and we already have a profile, the profile creation screen should not appear


app/src/main/java/com/android/sample/model/profile/ProfilesRepositoryFirestore.kt
```diff
@@ -7,7 +7,7 @@ import com.google.firebase.firestore.FirebaseFirestore
 class ProfilesRepositoryFirestore(private val db: FirebaseFirestore) : ProfilesRepository {
 
   override fun getUser(userId: String, onSuccess: (User?) -> Unit, onFailure: (Exception) -> Unit) {
-    db.collection("profiles").document("profiles-$userId").get().addOnCompleteListener { task ->
+    db.collection("profiles").document(userId).get().addOnCompleteListener { task ->
```
GhaliBerbich, line 10: Good job on going back to the first implementation we had to fetch documents in the db.

GhaliBerbich (commented):
This change of logic seems good to me.

PR #40, opened 2024-10-14
feature(profile) : ui elements for the navigation of profileScreen added

GhaliBerbich:
added the bottomBar, and a topAppBar, added the navigation in the profile, as well as in the test. Only small changes were made in mainActivity to accomodate with the navigation implemented in the rest of the code.

### DIFF
--- a/app/src/androidTest/java/com/android/sample/ui/profile/ProfileScreenTests.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/profile/ProfileScreenTests.kt
@@ -4,8 +4,11 @@ import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertTextEquals
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.performClick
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/android/sample/ui/profile/Profile.kt
aminebenji04, line 1: Thanks for this! The missing navigation actions were obligating me to rerun the app everytime!
We maybe should try to find a way to not copy-paste the structure of the bottomNavigationMenu as most of its parameters stay the same

app/src/main/java/com/android/sample/MainActivity.kt
aminebenji04, line 1: Nothing to say, LGTM!

app/src/androidTest/java/com/android/sample/ui/profile/ProfileScreenTests.kt
aminebenji04, line 1: Maybe you can still add new tests, with for example testing the behavior of the navigationAactions, the goBackButton, the edit and add info button, the case where we put wrong info or good info

aminebenji04 (approved):
LGTM! We just have to implement more tests, you can see the detailed comments I gave youfor each file

PR #42, opened 2024-10-15
Add the tests and dialogs 

Yassyass181104:


elvan03:
The code overall looks great, thanks a lot for your work!
Small improvement suggestions for the CreateActivity class could be: 
-I see the catch block for NumberFormatException is empty (also in EditActivity). You could add some logging or a comment indicating why you're catching that exception. 
-The top part of CreateActivity code is getting a bit long. You could move all the OutlinedTextField() (the input fields section) into a helper function for better readability. 
Overall LGTM!

PR #44, opened 2024-10-15
Refactor/create activites

elvan03:
I took off the go back button in CreateActivity screen and added a bottom bar (since it's a main screen accessible from bottom app bar). I also added the scroll option. 
I put all strings in the string.xml file for better readability & maintainability.

Solves #43


Yassyass181104 (commented):
This pull request effectively adds startTime and duration to the Activity model and integrates them into UI screens like CreateActivityScreen and EditActivityScreen. The feature implementation is solid, but there are areas for improvement, such as reducing code duplication by creating reusable input components, handling errors when converting string inputs to numbers, and ensuring consistent UI spacing for better readability. Default values should be centralized to avoid hardcoding, and unnecessary nesting in the NavHost should be removed for cleaner navigation logic.

app/src/main/java/com/android/sample/model/activity/ActivitiesRepositoryFirestore.kt
```diff
@@ -38,6 +38,8 @@ class ActivitiesRepositoryFirestore(private val db: FirebaseFirestore) : Activit
                       data["title"] as String,
                       data["description"] as String,
                       data["date"] as Timestamp,
+                      data["startTime"] as? String ?: "00:00",
```
DAFERMOH, line 41: good implementation but could improve data["startTime"] by giving it a Timestamp.

app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
```diff
@@ -131,6 +134,22 @@ fun ActivityDetailsScreen(
             Spacer(modifier = Modifier.width(4.dp))
             Text(text = dueDate ?: "not defined yet")
           }
+/*
```
DAFERMOH, line 137: maybe this part could be merged in a single row or keep it this way and use different icons or else the UI will not look good.

DAFERMOH (commented):
good job, you could just take into account some improvements that I mentionned in your PR.

PR #48, opened 2024-10-16
refactor(signIn): refactor signIn to be implemented following the MVVM model in order to facilitate testing process.

aminebenji04:
- Created new filesSignInViewModel.kt and SignInRepository.kt and SignInRepositoryFirebase.kt to follow the MVVM structure done with the other features

- Removed the business logic functions from the SignIn.kt file to move them into the view model.

- Added a factory object for the SignInViewModel to follow what we already done with the other view models.


app/src/main/java/com/android/sample/ui/authentication/SignIn.kt
Yassyass181104, line 1: The code looks well improved and way cleaner great job

Yassyass181104 (approved):
Overall : Well done, the code looks shorter and optimized. As I can see the limit cases are handled better than before, well done

PR #53, opened 2024-10-16
Feature/add activity linking

GhaliBerbich:


app/src/androidTest/java/com/android/sample/ui/activity/CreateActivityScreenTest.kt
aminebenji04, line 1: I think you can merge all these tests in the same tests as they test the same thing for different components, you can name it: fun assertAllComponentsAreDisplayed for example.

app/src/main/java/com/android/sample/model/profile/ProfileViewModel.kt
aminebenji04, line 1: Thank you for this, this was what we needed, the problem was that we could only fetch all the activities in one shot, now we can access each activity specifically.

app/src/main/java/com/android/sample/model/profile/ProfilesRepositoryFirestore.kt
aminebenji04, line 1: LGTM! Thanks to the listener you added, we will see dthe updates in live in the profile

app/src/main/java/com/android/sample/ui/activity/EditActivity.kt
aminebenji04, line 1: No notable changes that could mess up with the tests

app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
aminebenji04, line 1: LGTM

aminebenji04 (approved):

PR #54, opened 2024-10-16
Feature/list activities in profile

hmaryam84:
### 1. Lists of Activities in Profile Section
* Added two lazy columns in the **Profile** section:
  * A list of activities created by the logged-in user.
  * A list of activities the user is enrolled in.
* Clicking on a created activity navigates the user to the **Edit Activity** screen.
* Clicking on an enrolled activity navigates the user to the **Activity Details** screen.

### 2. Activity Types: PRO, INDIVIDUAL, SOLO
* Each activity now has a type: `PRO`, `INDIVIDUAL`, or `SOLO`. (These types are explained in the project's wiki).
* Users can choose the type when creating an activity.
* The activity type can also be modified when editing the activity.
* In the **Overview** screen, users can filter activities by type:
  * Show all activities.
  * Filter by a specific activity type.

### Tests
* I acknowledge the lack of tests for these features.
* Due to urgency, I am pushing these changes so other team members can integrate them.
* I will add tests in an upcoming commit.


elvan03:
The code looks great! Thanks for explaining so clearly the changes you made.

A few things I noticed that could be "improved":
For EditProfile, the code for handling the case where the profile is not selected "No profile selected. Should not happen" is repeated 2 times. You could use a function for error messages or just put it in the string.xml file (more generally all strings could be in the string.xml file).
In Profile's screen, it could be nice to display a message like "no activities available" for when there are no activities(for example right below "Activities Enrolled in"), a bit like you did ViewListActivities with “There is no activity yet.” 

Otherwise LGTM, thanks!

PR #55, opened 2024-10-17
feat(profile management): Add functions to update profile and delete …

GhaliBerbich:
…activity from profile

Implement new functions in the ProfileViewModel to support updating the user profile and deleting an activity by its ID. These functionalities extend the ViewModel's capabilities to handle user-specific operations directly from the UI. Corresponding methods have been added to the FirestoreRepository to carry out the actual database operations. Additionally, necessary signatures for these new functions have been introduced in the ProfilesRepository interface to maintain consistency and ensure implementation across any future repositories.

### DIFF
--- a/app/src/main/java/com/android/sample/model/profile/ProfileViewModel.kt
+++ b/app/src/main/java/com/android/sample/model/profile/ProfileViewModel.kt
@@ -52,6 +52,18 @@ open class ProfileViewModel(private val repository: ProfilesRepository, userId: 
         onFailure = {})
   }
 
+  fun updateProfile(user: User) {
+    repository.updateProfile(user = user, onSuccess = { fetchUserData(user.id) }, onFailure = {})
+  }
+
+  fun deleteActivityFromProfile(userId: String, activityId: String) {
+    repository.deleteActivityFromProfile(
+        userId = userId,
+        activityId = activityId,
+        onSuccess = { fetchUserData(userId) },
+        onFailure = {})
+  }
+
   companion object {
     fun Factory(uid: String): ViewModelProvider.Factory =
         object : ViewModelProvider.Factory {
--- a/app/src/main/java/com/android/sample/model/profile/ProfilesRepository.kt
+++ b/app/src/main/java/com/android/sample/model/profile/ProfilesRepository.kt
@@ -10,5 +10,14 @@ interface ProfilesRepository {
       onFailure: (Exception) -> Unit
   )
 
+  fun updateProfile(user: User, onSuccess: () -> Unit, onFailure: (Exception) -> Unit)
+
   fun addProfileToDatabase(userProfile: User, onSuccess: () -> Unit, onFailure: (Exception) -> Unit)
+
+  fun deleteActivityFromProfile(
+      userId: String,
+      activityId: String,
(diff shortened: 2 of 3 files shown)


app/src/main/java/com/android/sample/model/profile/ProfilesRepositoryFirestore.kt
```diff
@@ -92,4 +93,47 @@ open class ProfilesRepositoryFirestore(private val db: FirebaseFirestore) : Prof
...
+      if (result.isSuccessful) {
+        onSuccess()
+      } else {
+        result.exception?.let { e ->
+          Log.e("TodosRepositoryFirestore", "Error performing Firestore operation", e)
+          onFailure(e)
+        }
+      }
+    }
+  }
```
mohamedtahaguelzim, line 138: why did you create this function ? I do not think it is a good idea, as it generalises the log message, which would have been useful to have personnalised, as in line 116.

Plus, if you created this function why don't you use it in deleteActivityFromProfile function ?

app/src/main/java/com/android/sample/model/profile/ProfilesRepository.kt
aminebenji04, line 1: can't we find the activty Id from the profile Id or the inverse?

app/src/main/java/com/android/sample/model/profile/ProfileViewModel.kt
aminebenji04, line 1: LGTM

mohamedtahaguelzim (changes requested):
A decision is needed for performFirestoreOperation.

If you keep it, then use it for all functions,
else I believe you should remove it.

app/src/main/java/com/android/sample/model/profile/ProfilesRepository.kt
GhaliBerbich: yes but we want to delete a specific activity from all the activities the profile has. So it is to specify which one has to be deleted.

mohamedtahaguelzim:
These should have been pushed with the other PR that needed those functions.

Also, in a more global way, I believe that deleteActivityFromProfile has no reason to exist, as it is a consequence of deleting an activity or un-enrolling from an activity, thus we need to implement it in the deleteActivity or unenroll functions

But for the moment, the PR is very useful as it will fix the compiling issues we currently have in dev branch.

Thanks for the implementation.

app/src/main/java/com/android/sample/model/profile/ProfilesRepositoryFirestore.kt
GhaliBerbich: I decided to remove the helper for the moment, because it allows to put a specific log in each function, enabling us to know which operation failed. Using a helper, meant having the same log for all operations.

app/src/main/java/com/android/sample/model/profile/ProfilesRepositoryFirestore.kt
mohamedtahaguelzim, line 1: Great that logging is now personnalised. We can add that helper function later, if we find it really useful.

Thanks!

mohamedtahaguelzim (approved):
Approved for me! Thanks for the changes.

PR #63, opened 2024-10-18
Feature/editing profile

GhaliBerbich:
I only changed the files in the commits about the profile editing. It enables to modify the infos a profile apart the photo as we still have not implemented the possibility to upload a photo in the app.


Yassyass181104 (approved):
Changes are coherent some tests on the CI doesn't work but it's going to be fixed in the next PR, Great!

PR #66, opened 2024-10-18
Feature/enroll image

DAFERMOH:
Added tests and fixed the display of selectedActivity's data (I had to call select activity on ViewListActivities.kt) since some merge deleted it.


elvan03:
LGTM!

PR #71, opened 2024-10-18
Feature/enroll image

DAFERMOH:
fixed ActivityDetailsScreen and it's test to work on small phone by adding the ability to scroll
NB: if you see a lot of modifications in my files it's due to ktfmtFormat.


aminebenji04:
Small changes, nothing to say !

PR #73, opened 2024-10-18
refactor(LoadingScreen): Redirect guests to Signup screen

GhaliBerbich:
Refactored the LoadingScreen to redirect guests directly to the Signup screen. Implemented navigation logic and updated UI components to ensure a seamless transition for new users.


aminebenji04:
LGTM

PR #77, opened 2024-10-28
build(dependencies): add dependencies for kapt and hilt.

aminebenji04:


Yassyass181104 (approved):
Thank you, those changes are necessary for the implementation of DI in the project

PR #83, opened 2024-10-30
Fix/profile navigation guest mode

GhaliBerbich:
Also put a line so that the profile screen appears after the first time we launch the app and create the profile. Before, we had to launch the app, create the profile, then to see the profile we had to close the app and launch it again and only from that point, the profile appeared.


aminebenji04 (approved):
Thanks for finding this bug, it was scaring me! LGTM !

PR #84, opened 2024-10-30
Fix/fix app crashes

Yassyass181104:
Fixed the problem that made the app crash when creating an activity. Deleted the places left input space and instead just converted it to total places and used the size of the list of participants. Also added the possibility to retrieve the participants from the repository


aminebenji04 (approved):
It looks good. I don't think that we need more tests on this. We still have to check that there are no edge cases that can make our app crash

PR #92, opened 2024-10-31
Fix/profile creation sync viewmodel

GhaliBerbich:


aminebenji04 (approved):
Thanks, fetching the user data when creating a profile will fix the issue of empty profile when creating a profile and going to the profile screen.

PR #94, opened 2024-10-31
Map backend

elvan03:
I added the MVVM structure for the map feature. 
I created:
-Location data type
-Location repository
-LocationViewModel, in order to retrieve location suggestions based on a user query (coming in UI later)
-NominatimLocationRepository, in order to (later) convert string locations into lat & long coordinates.

I also adapted my code to use Hilt annotations for providing dependencies and code modularity.
Code coverage > 80 %
This is the first step in order to implement the map, the front-end is coming in a later PR.

This solves issue #68


Yassyass181104 (approved):
Those new files are what we wanted, perfect !

aminebenji04:
Code coverage looks great, let's play around with the app to see if there is no hidden bug

PR #95, opened 2024-10-31
Feature/auth/password field

aminebenji04:
I added the feature that was asked: being able to hide the password when writing it along with the tests


Yassyass181104 (approved):
It fixes the bug we had with authentication, very good

elvan03:
LGTM!

PR #100, opened 2024-10-31
fix:fix creator is a participant

Yassyass181104:
Simple change to add the owner of the activity to the activity participants


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -308,6 +308,11 @@ fun CreateActivityScreen(
                 val calendar = GregorianCalendar()
                 val parts = dueDate.split("/")
                 if (parts.size == 3 && timeFormat.size == 2 && durationFormat.size == 2) {
```
DAFERMOH, line 310: it's good that you thought about adding the user the list of participants once you enroll in an activity.

DAFERMOH:
LGTM

DAFERMOH (approved):

PR #104, opened 2024-10-31
view enrolled people and add location, starting time and duration on the Activity Details Screen and link it to the actual activity's arguments

DAFERMOH:
view people enrolled in an activity in ActivityDetailsScreen and add start time and duration in ActivityDetailsScreen and fixed issue that allowed a user to enroll twice


aminebenji04 (approved):
LGTM You just have to change the name of the test file because putting numbers is not a good practice. Overall it looks great Iand I can't see where a problem could happen

PR #105, opened 2024-10-31
refactor(di profile): add dependency injection in profileViewModel

GhaliBerbich:
changed profileViewModel signature because a parameter was never used, changed the main to use the dependency injection for the ProfileViewModel.


hmaryam84 (commented):
add dependency injection for Profile view model. This looks correct, you didnt forget to change the main and add the gilt viewModel for the profileViewModel. Good!

aminebenji04 (approved):
LGTM! This is some classic DI implementation.

PR #111, opened 2024-11-01
add list of participants to ActivityDetailsScreen

DAFERMOH:
add the ability to see people enrolled in an activity in activityDetailsScreen.


aminebenji04 (approved):
It looks good now, we have to see in-app to see if nothing is missing.
What I already see now is that you can add the profile picture of the participants

PR #120, opened 2024-11-03
Feature/auth/already logged in

aminebenji04:
This PR introduces a new Choose Account Screen that displays when a user is already logged in. Like in Instagram or Facebook, the screen gives the option to either continue with the current account or switch to a different account.

Key changes include:

- Choose Account Screen: A new screen that appears if the user is already logged in, allowing the user to select their current account or switch accounts.

- Profile Picture Integration: Displays the user's profile picture on the Choose Account Screen.

- Sign-Out Functionality: A  function  in the SignInViewModel, simplifying the process of signing out from different parts of the application.

- Navigation Actions: Updates to the navigation structure to incorporate the new Choose Account Screen.

### DIFF
--- a/app/src/main/java/com/android/sample/MainActivity.kt
+++ b/app/src/main/java/com/android/sample/MainActivity.kt
@@ -1,6 +1,7 @@
 package com.android.sample
 
 import android.os.Bundle
+import android.util.Log
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.layout.fillMaxSize
@@ -23,6 +24,7 @@ import com.android.sample.resources.C
 import com.android.sample.ui.activity.CreateActivityScreen
 import com.android.sample.ui.activity.EditActivityScreen
 import com.android.sample.ui.activitydetails.ActivityDetailsScreen
+import com.android.sample.ui.authentication.ChooseAccountScreen
 import com.android.sample.ui.authentication.SignInScreen
 import com.android.sample.ui.authentication.SignUpScreen
 import com.android.sample.ui.listActivities.ListActivitiesScreen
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/android/sample/MainActivity.kt
Yassyass181104, line 1: You managed smoothly how to choose either there is already a user

app/src/main/java/com/android/sample/model/auth/SignInViewModel.kt
Yassyass181104, line 1: Good management of the signout procedure

app/src/main/java/com/android/sample/ui/authentication/ChooseAccount.kt
Yassyass181104, line 1: Nothing to say looks good

Yassyass181104 (approved):
Overral it's good even if I don't understand very well why do you use item on the lazy element apart that everything is good

app/src/main/java/com/android/sample/model/auth/SignInViewModel.kt
elvan03, line 1: Since you use dependency injection, I guess you don't need the factory anymore and you could remove it.

app/src/main/java/com/android/sample/ui/authentication/ChooseAccount.kt
elvan03, line 1: Code looks great! you could avoid hardcoding the space values or the colours and take them from theme instead, but it's totally acceptable since we'll take care of design in the coming weeks.

elvan03 (approved):
Overall looks great! You could look into removing the factory in the SignInViewModel as you use dependency injection now.
Otherwise good job!

Issue #121, opened 2024-11-03
create map screen and add it in the navigation


elvan03:
This task is about setting up the map api key and configure gradle files in the project. Then we should also be able to see the map displaying as a main screen of the app

PR #122, opened 2024-11-03
Map screen

elvan03:
This PR contains the map, added as a main screen. I set-up and added the API key (for using google maps) in the gradle.properties file, which is now in the gitignore. 

PS: the test coverage is not above 80% but this is hard to achieve as right now the map screen is very simplistic. In my next task, I will be adding a lot more to the map screen and the test coverage should then pass the required threshold.

PS: I unfortunately pushed the gradle.properties file :/
To solve this, I have changed our actual API key so no worries :)

Solves issue #121


.gitignore
aminebenji04, line 1: Nevermind

app/build.gradle.kts
aminebenji04, line 1: Looks like the bootcamp, no problem for me here!

app/src/main/AndroidManifest.xml
aminebenji04, line 1: Ok!

app/src/main/java/com/android/sample/MainActivity.kt
aminebenji04, line 1: Nothing to say, classic screen adding

app/src/main/java/com/android/sample/ui/navigation/NavigationActions.kt
aminebenji04, line 1: Nothing to say, classic screen adding

app/src/main/java/com/android/sample/ui/map/Map.kt
aminebenji04, line 1: Looks good to me, this will juste display the map without anything on it, so I don't see any problems here!

aminebenji04 (approved):
Everything looks good, this is a Map setup and there is no obvious problem source. This is clean !

PR #124, opened 2024-11-04
I just changed the build dependencies for it to work in all our computers

aminebenji04:


Yassyass181104 (approved):
Yes, you upgraded almost everything so it works for everyone. Great (Don't forget to solve mege conflicts)

mohamedtahaguelzim (approved):
Thanks for providing this short and clear fix. It is unfortunate that we require a custom test runner to fix our issue. Most important is that the CI passes. All looks good to me. Thanks!

PR #130, opened 2024-11-04
Addition to PR #122

elvan03:
This PR is to add the map in the bottom bar. It was already done but probably was deleted in a previous merge so I added it again.

Addition to PR 122 that solves issue #121


aminebenji04 (approved):
Nothing to say!

PR #158, opened 2024-11-09
Fix (profile creation): no profile creation without name and surname

aminebenji04:
In this PR:
- I added the TextFieldWithErrorState composable, a reusable component designed for input validation with error handling. 
- I fixed the issue of being able to create a profile without a name and a surname using this new component
- I used the new component in the sign in
- I updated some old tests where I was clicking on the "create profile" button without filling the name and the surname

### DIFF
--- a/app/src/main/java/com/android/sample/ui/authentication/SignIn.kt
+++ b/app/src/main/java/com/android/sample/ui/authentication/SignIn.kt
@@ -82,7 +82,7 @@ fun SignInScreen(navigationActions: NavigationActions, viewModel: SignInViewMode
                     if (input.isBlank()) "Password cannot be empty" else null
                   },
                   externalError = passwordErrorState.value,
-                  errorTestTag = "PasswordError")
+                  errorTestTag = "PasswordErrorText")
 
               Spacer(modifier = Modifier.height(16.dp))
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/android/sample/ui/authentication/SignIn.kt
Yassyass181104, line 1: As I can see you added the fact that password can't be empty which seems to solve an issue, very good job. You changed the code in a way that it signals an error if it's empty

app/src/main/java/com/android/sample/ui/components/Textfields.kt
Yassyass181104, line 1: Here you created the composable you used to implement the error signalement for password maybe adding a regex or minimal size can be very good too

app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
Yassyass181104, line 1: Good solution to solve the problem of being able to create an "empty" profile, please try to avoid putting commented code on the main.

Yassyass181104 (approved):
Overral very good, it solves a problem we had the code is well modularized and clean the UI seems to be more familiar with what we have in other apps and websites. Except some imperfections that I pointed out in my comments, well done !

PR #159, opened 2024-11-09
Remove the white gap for a better ui

hmaryam84:
* remove the gap between the list of activities and the bottom bar
 * remove the white gap between the list of activities and the filtering bar

(This is a task I had 2 sprints ago, but I pushed the bad code, now the issue is fixed)


Yassyass181104 (approved):
This is a good solution, the only problem which is not really a problem is that under the bottombar there will be activities that are displayed but will not appear. A solution would be considering the row you use as a filter as the bottombar, keep the lazy column and the bottombar as they were. But your solution is equally valid, great job

PR #161, opened 2024-11-10
fix: reset profile data on logout to prevent residual user information

aminebenji04:
In this PR, I just added a new function in the profileViewModel and called it in the profileScreen, you can see the details in my commit messages.
### How to try this new fix:
1. Log-in with you account or with Google
2. Go to the profile screen 
3. Click on the "..." dots on the top right corner and then click on "Logout"
4. Log-in as a Guest and go to the profile screen
5. Now you should not have the previous profile showing


elvan03 (approved):
Looks clean!

Yassyass181104 (approved):
Looks like you nulify the profile which is good, great job

PR #162, opened 2024-11-10
Refactor/uncomment the profile screen tests

elvan03:
I created this PR in order to uncomment the test I had forgotten to uncomment in a previous PR. I also adapted the test to receive a location parameter of type Location instead of string.


Yassyass181104 (approved):
LGTM

PR #163, opened 2024-11-10
Fix/fix ambiguity users

Yassyass181104:
Apported modifications to solve the ambiguity between user and simple user so now we can add people that don't have a profile and enroll ourselfs having profiles. I did this by refactoring every test and class that included simpleUser and adapted it to be a user with null/empty arguments except the name and surname.


elvan03 (approved):
Changes make sense.
LGTM!

PR #166, opened 2024-11-11
Disable line break create and edit activity

hmaryam84:
## Disallow multiple lines in text fields in create and edit activity

 * When creating or editing an activity, only allow multiple lines for description
 * All the other text fields have a single line
 * This is for a better UX

### Remarks:
- There are no tests since I only modified the attributes of the text fields.
- The original issue was only about making these changes in Create Activity, but for more coherence, I also made the same changes is Edit Activity.


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
aminebenji04, line 1: This is a good idea, I will do the same for the files I created, profileCreation, signIn and signUp

aminebenji04 (approved):
Good fix, we should all do this

PR #171, opened 2024-11-11
fix: interests not being updated when creating a profile

aminebenji04:


app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
Yassyass181104, line 1: LGTM

Yassyass181104 (approved):
Great to solve

PR #172, opened 2024-11-11
Restructure/restructure tests dummy data

Yassyass181104:
In this PR I implemented the use of dummy data following these steps:

- I first created a file named DummyData where I store the data for the tests 
- Secondly I adapted the AndroidTests to use the data stored in DummyData
- And finally modified the unit tests to use the data stored in DummyData

By that, now we can use the same data for tests and cleans up the code


elvan03 (approved):
Nothing to say apart that this must be the most satisfying PR to merge! Dummy class looks correctly adapted to previous code, thanks!

aminebenji04 (approved):
Very good, even if you changed a lot of files, this is just optimization, so as you changed only tests and tests pass, there is nothing to say. LGTM!
What I can add is that I would like to know if the lines of code in DUmmyData are covered or not

PR #173, opened 2024-11-11
Selected activity not updated

hmaryam84:
### Select the activity when click on an enrolled activity in profile sreen

  * select the activity clicked on before navigating to its details screen
  * before the fix: the correct activity was not selected,
  resulting on the navigation to another activity details


elvan03 (approved):
Thanks for explaining the bug precisely and good job fixing it!

LGTM!

PR #174, opened 2024-11-11
Fix/deleted liked activities crash

hmaryam84:
### deal with deleted liked activities

# Explanation of the bug

 * When the user wants to go to the liked activities screen,
if one of the activities he has liked before was deleted by its creator, 
the app crashes
# Proposed solution
  * don't show deleted activities in liked activities
  * remove deleted activities from the liked activities list of the
   user when he tries to go on the like activities screen


app/src/main/java/com/android/sample/ui/listActivities/LikedActivities.kt
```diff
@@ -143,7 +143,11 @@ fun ActivityCard2(
     allActivities: List<Activity>,
 ) {
 
-  val activity = allActivities.filter { act -> act.uid == activityId }[0]
+  val activity = allActivities.find { act -> act.uid == activityId }
```
elvan03, line 146: Nice catch for using find!

You can further improve conciseness by handling the null case with the "?:" as follows:
 allActivities.find { it.uid == activityId } ?: run {...}

elvan03 (approved):
Suggested one small conciseness improvement, otherwise well done for catching why deleted liked activities crashed and finding a smart solution!

LGTM!

PR #178, opened 2024-11-12
Fix: switch the two buttons to switch account and to continue with logged account

aminebenji04:
In this PR:
- I just changed the order of the two buttons
- I made the bigger button be for continuing and the smaller for switching account


elvan03 (approved):
LGTM!

PR #179, opened 2024-11-12
Refactor/activity details screen

### LINKED ISSUE #134: Redundant time in activity detail screen
image
The time is written in the two fields. It should be removed from the first one.

elvan03:
In this PR, I solve the confusion about the "time displaying twice". Actually, one of them is the starting time of activity while the other is the duration of the activity. 

In order to solve the confusion, I changed the icon to a Timelapse icon that represents in my opinion the concept of length better. I also added a string: "Event length:" (which isn't the best visually) but right now we are sure there is no confusion. We could remove this when we find a better alternative.

I also took the chance to display the location field better, which used to be on the same line as the price field. It is now below it and in my opinion is a lot better and easier to read.

closes #134


aminebenji04 (approved):
LGTM, nothing to be afraid of, UI enhancements and tests working

PR #180, opened 2024-11-13
Fix/invalid time

### LINKED ISSUE #165: Activity time can be set to invalid value (e.g. 60:124)
When creating an activity, one can set any time for the activity, without restriction for invalid time values ( e.g. negative hour, time > 23:59, ...)

Bug discovered by @hmaryam84

hmaryam84:
### fix: invalid start time format in edit activity and create activity

# Explanation of the bug

 * it was possible to create activities with invalid start time such as: 65:80 or aa:bb.

# Proposed Solution

 * add the following start time conditions:
   * the hours should be between 00 and 23
   * the minutes should be between 00 and 59
   * there should be no letter
 * show a toast indicating invalid format of the start time when editing the activity using an invalid start time

 * the same changes were made in both edit activity and create activity for a coherent UX.
# Testing
 * no tests were added in this pr. I am waiting for other members to set up hilt so I could write end to end test.
  
  closes #165

### DIFF
--- a/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
+++ b/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
@@ -381,7 +381,9 @@ fun CreateActivityScreen(
                 enabled = title.isNotEmpty() && description.isNotEmpty() && dueDate.isNotEmpty(),
                 onClick = {
                   val timeFormat = startTime.split(":")
-                  if (timeFormat.size != 2) {
+                  val timePattern = Regex("^([01]?[0-9]|2[0-3]):[0-5][0-9]$")
+
+                  if (timeFormat.size != 2 || !timePattern.matches(startTime)) {
                     Toast.makeText(
                             context, "Invalid format, time must be HH:MM.", Toast.LENGTH_SHORT)
                         .show()
@@ -394,7 +396,7 @@ fun CreateActivityScreen(
                   }
                   val calendar = GregorianCalendar()
                   val parts = dueDate.split("/")
-                  if (parts.size == 3 && timeFormat.size == 2 && durationFormat.size == 2) {
+                  if (parts.size == 3 && timeFormat.size == 2 && durationFormat.size == 2 && timePattern.matches(startTime)) {
                     attendees += profileViewModel.userState.value!!
                     try {
                       calendar.set(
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -381,7 +381,9 @@ fun CreateActivityScreen(
                 enabled = title.isNotEmpty() && description.isNotEmpty() && dueDate.isNotEmpty(),
                 onClick = {
                   val timeFormat = startTime.split(":")
-                  if (timeFormat.size != 2) {
+                  val timePattern = Regex("^([01]?[0-9]|2[0-3]):[0-5][0-9]$")
+
+                  if (timeFormat.size != 2 || !timePattern.matches(startTime)) {
                     Toast.makeText(
```
mohamedtahaguelzim, line 387: It is a great thing that you chose to use a pattern for time. These a more standard and have for guarantee on the checks. Well done.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -394,7 +396,10 @@ fun CreateActivityScreen(
                   }
                   val calendar = GregorianCalendar()
                   val parts = dueDate.split("/")
-                  if (parts.size == 3 && timeFormat.size == 2 && durationFormat.size == 2) {
+                  if (parts.size == 3 &&
+                      timeFormat.size == 2 &&
+                      durationFormat.size == 2 &&
+                      timePattern.matches(startTime)) {
```
mohamedtahaguelzim, line 402: Again, good to use pattern matching.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
mohamedtahaguelzim: The pattern seems appropriate for the time use case. It is great that it blocks times above 23:59 (such as 27:91).

mohamedtahaguelzim (approved):
Thanks for using a pattern to check the time. It results in a clear check of the user input time. I approve these changes.

aminebenji04 (approved):
Good, in the future we can add tests that check how are handled the cases where we put the time in wrong/good format

PR #187, opened 2024-11-13
Refactor the content of the file using the variables in dummy data

Yassyass181104:
There were some variables that could have been refactored in SigninViewModelTest to use the variables in dummy data. I just did it and corrected the data


aminebenji04 (approved):
Very good, this wy we have less magic values

PR #192, opened 2024-11-14
Fix map permission not appearing when launching the app

aminebenji04:
In this PR, I just moved the map permission checking to the Map screen instead of the mainActivity so that we are not asked about location when launching the app but when we go to the map.


Yassyass181104 (approved):
Good change, coherent and usefull. Nothing to say since it's already commented code and it has just been moved

PR #194, opened 2024-11-14
Refactor the UI files to avoid using magic numbers 

Yassyass181104:
I added constants in the C file.  Those new constants will help us have uniform apppearance in the UI and avoid using magic numbers. There is not a big difference in UI nonobstant the code looks cleaner.

### DIFF
--- a/app/src/main/java/com/android/sample/resources/C.kt
+++ b/app/src/main/java/com/android/sample/resources/C.kt
@@ -17,5 +17,43 @@ object C {
     val ERROR_TEXTFIELD_PADDING_START = 16.dp
     val ERROR_TEXTFIELD_PADDING_TOP = 5.dp
     val ERROR_TEXTFIELD_FONT_SIZE = 12.sp
+
+    const val STANDARD_PADDING = 8
+
+    const val BUTTON_WIDTH = 300
+
+    const val BUTTON_HEIGHT = 40
+
+    const val MEDIUM_PADDING = 16
+
+    const val LARGE_PADDING = 24
+
+    const val SMALL_PADDING = 4
+
+    const val TITLE_FONTSIZE = 24
+
+    const val SUBTITLE_FONTSIZE = 16
+
+    const val WHITE_COLOR = 0xFFFFFFFF
+
+    const val BLACK_COLOR = 0xFF000000
+
+    const val IMAGE_SIZE = 100
+
+    const val LARGE_IMAGE_SIZE = 200
+
+    const val SMALL_IMAGE_SIZE = 30
+
+    const val MEDIUM_IMAGE_SIZE = 64
+
+    const val BOTTOM_MENU_HEIGHT = 60
+
+    const val BOTTOM_CORNER = 50
+
+    const val TOP_TITLE_SIZE = 30
+
+    const val TEXT_FONTSIZE = 12
+
+    const val WIDTH_FRACTION= 0.8f
   }
 }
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/android/sample/ui/authentication/SignUp.kt
Yassyass181104, line 1: Good change, missed this one thanks for taking a close look and solving some errors

app/src/main/java/com/android/sample/resources/C.kt
mohamedtahaguelzim, line 1: Thanks Yassine for doing this effort. We just need to make sure that these values are adequate with our Figma.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -151,81 +155,94 @@ fun CreateActivityScreen(
                   },
               )
             }
-            Spacer(modifier = Modifier.height(8.dp))
+            Spacer(modifier = Modifier.height(STANDARD_PADDING.dp))
```
mohamedtahaguelzim, line 158: I don't know how good of a practice would it be to directly store the .dp in the C file.

So that we no longer need to add .dp after each value.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -151,81 +155,94 @@ fun CreateActivityScreen(
...
             OutlinedTextField(
                 value = title,
                 onValueChange = { title = it },
                 label = { Text("Title") },
-                modifier = Modifier.padding(8.dp).fillMaxWidth().testTag("inputTitleCreate"),
+                modifier =
+                    Modifier.padding(STANDARD_PADDING.dp)
+                        .fillMaxWidth()
+                        .testTag("inputTitleCreate"),
```
mohamedtahaguelzim, line 166: Would be great if we also centralize the testTags !

app/src/main/java/com/android/sample/ui/activity/EditActivity.kt
```diff
@@ -510,26 +530,3 @@ fun EditActivityScreen(
...
-//          price = 0.0,
-//          placesLeft = 0,
-//          maxPlaces = 0,
-//          creator = "Creator",
-//          status = ActivityStatus.ACTIVE,
-//          location = "Location",
-//          images = listOf(),
-//          participants = listOf()))
-//  EditActivityScreen(navigationActions = navigationActions)
-// }
```
mohamedtahaguelzim, line 535: Thanks for deleting this useless comment. Makes the coee cleaner.

app/src/main/java/com/android/sample/ui/authentication/SignIn.kt
```diff
@@ -67,8 +75,8 @@ fun SignInScreen(navigationActions: NavigationActions, viewModel: SignInViewMode
             Image(
                 painter = painterResource(id = R.drawable.aptivity_logo),
                 contentDescription = "App Logo",
-                modifier = Modifier.size(110.dp).testTag("AppLogo"))
-            Spacer(modifier = Modifier.height(48.dp))
+                modifier = Modifier.size(IMAGE_SIZE.dp).testTag("AppLogo"))
+            Spacer(modifier = Modifier.height((2 * LARGE_PADDING).dp))
```
mohamedtahaguelzim, line 79: I guess it would have been better to add some sort of EXTRA_LARGE_PADDING

mohamedtahaguelzim (approved):
Thank you Yassine for this very tedious PR. You spent lots of efforts on it, but it brings great value in terms of UI-stability for our project.
I approve this PR. However, app should be ran to take a quick look at if things remain consistent. Also, I believe there is a test that checks for a specific padding (related to Comments). I think it should be refactored too. I will probably consider this in one of my upcoming PRs.
Thanks Again!

PR #195, opened 2024-11-14
Fix/activity creation in guest mode

### LINKED ISSUE #113: Activity creation in guest mode
The app crashes when trying to create an activty in guest mode

mohamedtahaguelzim:
Hi,
Very quick PR to block guests from creating an activity. They are rather prompted to sign in or create an account.
This feature has been tested in the "Guests" end-to-end test.

Thank you!

Close #113


hmaryam84 (approved):
Good for renaming the test, it is now representative of its functionality.
However, clicking on create button as a guest is useless in this case, since a guest cannot create an activity, and you don't test that a tost is displayed precenting the guest from creating an activit. We can notice that the app does not crsh as it did few sprints ago.

PR #197, opened 2024-11-15
tests(viewModel): raise test coverage in profile view model and create new tests for image view model.

aminebenji04:


hmaryam84 (approved):
Thank you for cleaning the code and adding some tests, this should be very helpful for our test coverage!

PR #198, opened 2024-11-15
test: make all the tests in create activity run in logged in mode

hmaryam84:
* Since we have changed the conditions to be able to add an activity, the tests should not be run in guest mode or otherwise they will not pass
 * mock a profile to enable adding activities testing


mohamedtahaguelzim (approved):
Oh wow, thank you so much @hmaryam84 ...
This was what I was missing in my PR!

L-**very**-GTM

PR #205, opened 2024-11-15
add MAP_API_KEY in CI.yaml from github secrets

DAFERMOH:
just added MAP_API_KEY in CI.yaml from github secrets to build the apk with the correct value of MAP_API_KEY.


Yassyass181104 (approved):
Perfect, solves an important issue

PR #210, opened 2024-11-15
Test: Separate End to End

hmaryam84:
### test: separating end to end

 * separate one end to end into 2 end to ends
 * the first one verify that the guest can access the overview, see activities and activity details, he also can filter activities
 * the second one verifies:

      * that in the profile screen, the user does not have any profile but a button    that invites him to sign in
      * in the liked activities screen, the user does not have any liked activities but a button that invites him to log in
      * for the moment the guest can fill the fields in create activities but cannot create any


mohamedtahaguelzim (approved):
Thanks for splitting the first e2e to two different ones. They had differents flows and it makes much more sense now ! Thanks !

PR #211, opened 2024-11-15
fix: change ui layout of comment reply button

### LINKED ISSUE #154: Post Reply button does not appear
When writing a reply, the post button does not appear which disallows us to post the reply.

mohamedtahaguelzim:
* bad layout made the button disappear. Switch from Row to Column.

Close #154


hmaryam84 (approved):
Very simple pr, a small step in the code, a big step for the application :)
Now the reply button should appear!

PR #218, opened 2024-11-19
Feature/disenroll

### LINKED ISSUE #217: add the ability to leave an activity a user is enrolled in

elvankonukseven:
**_This PR_** allows users to leave activities they have enrolled in, as this was not possible before.
I have changed the logic in Activity details screen in order to have a “enroll” or “leave” button depending on if we are already enrolled or not. 
A toast displays when you "leave" an activity successfully. 
I also had to change the navigation (when you enrol in an activity you are brought back to overview or if you leave you are brought back to your profile and activity is removed from your profile). 
Thus I also changed a few things in the MVVM of the Profile, add new method like "removeJoinedActivity". 

**Structure of the profile MVVM:** 

If you look at my commits, I initially wanted to merge methods like removeEnrolled and removeLikedActivities (and others) from profile MVVM but realised it was better to keep them separate since the context of where they’re being called is different. Also, it feels more intuitive to keep them separate and easier to use since the function names are more representative of what they do. (+ a lot of code already depends on these methods)

I also thought a lot about making a distinction between created activities and the joined activities since right now they are both in 1 same argument, activities. But again, since a lot of code and logic on this has been implemented already, it's a pain to change everything + it felt like it would add complexity.

**Adapt tests from LocationViewModel to use Mockito instead of Mockk**

I changed all tests from the test class to use mockito instead of mockk because I thought my CI issue was coming from using different mocking libraries. Indeed, I had a big CI issue that in the end came from test interferences (one test didn't finish before CI passed to next one) and everything crashed. Thus issue didn't come from LocationViewModel but at least now we are not using the Mockk library at all, I also deleted mockk dependencies from the gradle and it might avoid issues in later tests or library incompatibilities.

**Tests for ActivtiyDetailsScreen**

I have added extensive tests for the part of code I added in ActivityDetailsScreen. The coverage on new code looks low but you can see all the tests I have added, some of them (like when I test the toast for example) weren't taken into account by sonar somehow, even though they do locally with Jacoco. Not sure the 41% here is representative. 

closes #217

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -257,13 +257,13 @@ dependencies {
     androidTestImplementation(libs.androidx.espresso.intents)
     androidTestImplementation(libs.androidx.ui.test.junit4)
     androidTestImplementation(platform(libs.androidx.compose.bom))
-    testImplementation(libs.mockito.core)
 
-
-    testImplementation(libs.mockito.inline)
+    // Added for consistent Mockito usage across tests
+    testImplementation(libs.mockito.core.v570)
     testImplementation(libs.mockito.kotlin)
-    androidTestImplementation(libs.mockito.android)
     androidTestImplementation(libs.mockito.kotlin)
+    androidTestImplementation("org.mockito:mockito-android:5.7.0")
+
     testImplementation(libs.robolectric)
     androidTestImplementation(libs.kaspresso)
     androidTestImplementation(libs.kaspresso.allure.support)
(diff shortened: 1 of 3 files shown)


aminebenji04 (approved):
LGTM! There is nothing bad here, maybe you can add more comments in the viewModel to help us understand better your code.
Also we should test this on the phone and see if all the edge cases are handled, for example, what happens when joining an activity, then quitting it, then joining it again

PR #223, opened 2024-11-19
fix(tests): remove the mock file of profile repository.

aminebenji04:
This pull request removes the unnecessary MockProfilesRepository file and updates the test cases to use a mocked ProfilesRepository instance created with Mockito. The change ensures that the test cases are cleaner and follow best practices by removing redundant files and simplifying the test setup.


mohamedtahaguelzim (approved):
Mocking in the unit tests is preferable to using a copy of the MockProfilesRepository class. Indeed, you only require very few functions of it. Great!

PR #231, opened 2024-11-22
Add Distance Calculation and Display Features

### LINKED ISSUE #229: Activity shows distance from user if they allowed localisation
- [ ] calculate distance from user
- [ ] show distance in Overview screen
- [x] show distance in Activity Details screen

mohamedtahaguelzim:
This PR introduces the ability to calculate and display the distance between the user's current location and activity locations. The changes include enhancements to both the logic and UI, as well as tests to ensure accuracy and robustness.

---

### Features
1. **Distance Calculation**:
   - Implemented `getDistanceFromCurrentLocation` in `LocationViewModel` to calculate the distance (in kilometers or meters) between the user's current location and a target location.
   - Used the Haversine formula for distance calculation instead of `android.location.Location` due to compatibility issues.
   - Enhanced behavior to return `null` instead of `0` when location parameters are `null`, avoiding misleading results.

2. **UI Integration**:
   - **ActivityCard**:
     - Added distance display in the `ActivityCard`, showing the distance in meters (m) or kilometers (km).
     - Integrated location fetching upon opening the overview screen, with user permission prompts.
   - **ActivityDetailsScreen**:
     - Displayed distance information using pre-fetched location data from the overview screen.

---

### Tests
- Added unit tests for:
  - Handling `null` cases when permissions are not granted or location data is unavailable.
  - Verifying the accuracy of distance calculations between two points (delta ≤ 0.05 km).
  - Ensuring the behavior of returning `null` for `null` parameters in `getDistanceFromCurrentLocation`.
  - Added 6 UI tests, to check if content is well rendered when in km, m, and null.
- Adapted old tests to support the added view model.

---

### Notes
- The distance display depends on user location permissions and assumes pre-fetched location data for certain screens (e.g., `ActivityDetailsScreen`).
- This PR does not include fallback mechanisms for denied location permissions but handles `null` gracefully.

---

This implementation enhances the user experience by providing relevant distance details for activities and ensures accuracy and robustness in calculations. This implementation will be useful for the sorting recommendation algorithm.

Closes #229

### DIFF
--- a/app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
@@ -344,7 +344,9 @@ class OverviewScreenTest {
           activity = activity,
           distance = 12.354f)
     }
-    composeTestRule.onNodeWithTag("distanceText").assertTextContains("Distance : 12.3 km")
+    composeTestRule
+        .onNodeWithTag("distanceText", useUnmergedTree = true)
+        .assertTextContains("Distance : 12.4km")
   }
 
   @Test
@@ -359,7 +361,7 @@ class OverviewScreenTest {
           activity = activity,
           distance = null)
     }
-    composeTestRule.onNodeWithTag("distanceText").assertIsNotDisplayed()
+    composeTestRule.onNodeWithTag("distanceText", useUnmergedTree = true).assertIsNotDisplayed()
   }
 
   @Test
--- a/app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
@@ -1,7 +1,9 @@
 package com.android.sample.ui.listActivities
 
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsNotDisplayed
 import androidx.compose.ui.test.assertIsSelected
+import androidx.compose.ui.test.assertTextContains
 import androidx.compose.ui.test.hasText
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
@@ -10,12 +12,16 @@ import androidx.compose.ui.test.performClick
 import androidx.compose.ui.test.performScrollToNode
 import androidx.compose.ui.test.performTextClearance
 import androidx.compose.ui.test.performTextInput
+import androidx.test.rule.GrantPermissionRule
 import com.android.sample.model.activity.ActivitiesRepository
 import com.android.sample.model.activity.Activity
 import com.android.sample.model.activity.ActivityStatus
 import com.android.sample.model.activity.ActivityType
 import com.android.sample.model.activity.ListActivitiesViewModel
 import com.android.sample.model.map.Location
+import com.android.sample.model.map.LocationRepository
+import com.android.sample.model.map.LocationViewModel
+import com.android.sample.model.map.PermissionChecker
 import com.android.sample.model.profile.ProfileViewModel
 import com.android.sample.model.profile.ProfilesRepository
 import com.android.sample.model.profile.User
@@ -40,6 +46,7 @@ class OverviewScreenTest {
   private lateinit var navigationActions: NavigationActions
   private lateinit var listActivitiesViewModel: ListActivitiesViewModel
   private lateinit var userProfileViewModel: ProfileViewModel
+  private lateinit var locationViewModel: LocationViewModel
   private lateinit var testUser: User
 
   private val activity =
@@ -79,6 +86,13 @@ class OverviewScreenTest {
 
   @get:Rule val composeTestRule = createComposeRule()
 
+  @get:Rule
+  val permissionRule: GrantPermissionRule =
+      GrantPermissionRule.grant(
+          android.Manifest.permission.ACCESS_FINE_LOCATION,
+          android.Manifest.permission.ACCESS_COARSE_LOCATION,
+          android.Manifest.permission.CAMERA)
+
   @Before
   fun setUp() {
     profilesRepository = mock(ProfilesRepository::class.java)
@@ -101,6 +115,8 @@ class OverviewScreenTest {
     // `when`(userProfileViewModel.userState).thenReturn(MutableStateFlow(testUser))
 
     `when`(navigationActions.currentRoute()).thenReturn(Route.OVERVIEW)
+    locationViewModel =
+        LocationViewModel(mock(LocationRepository::class.java), mock(PermissionChecker::class.java))
   }
 
   @Test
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/android/sample/MainActivity.kt
hmaryam84, line 1: Here you just added the locationViewModel. I understand its usefulness as you use it for the distance calculations. Great.

app/src/main/java/com/android/sample/model/map/LocationViewModel.kt
```diff
@@ -4,6 +4,7 @@ import android.util.Log
 import androidx.lifecycle.ViewModel
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
+import kotlin.math.*
```
hmaryam84, line 7: Importing the **whole** math library is useful here as you use 6 different math functions. If you used less, I'd have recommended to only import the necessary functions.

app/src/main/java/com/android/sample/model/map/LocationViewModel.kt
```diff
@@ -24,6 +25,10 @@ constructor(
   private val _currentLocation = MutableStateFlow<Location?>(null)
   val currentLocation: StateFlow<Location?> = _currentLocation
 
+  fun setCurrentLocation(location: Location?) {
```
hmaryam84, line 28: I will check for the way you manage the null cases, because you seem to willingly allow it here... Careful with this !!

app/src/main/java/com/android/sample/model/map/LocationViewModel.kt
```diff
@@ -55,4 +60,51 @@ constructor(
...
+    val dLat = lat2Rad - lat1Rad
+    val dLon = lon2Rad - lon1Rad
+
+    val a = sin(dLat / 2).pow(2) + cos(lat1Rad) * cos(lat2Rad) * sin(dLon / 2).pow(2)
+
+    val c = 2 * atan2(sqrt(a), sqrt(1 - a))
+
+    // Distance in kilometers
+    return earthRadius * c
+  }
```
hmaryam84, line 109: Thanks for providing the comments documentations. The functions speak for themselves. Weird how you couldn't use android.location.Location distanceTo function! But I guess it is okay, as you used an official calculation function. The math implementation seems correct. I just compared it with the original formula.

app/src/main/java/com/android/sample/model/map/LocationViewModel.kt
```diff
@@ -55,4 +60,51 @@ constructor(
...
     }
   }
+
+  /**
+   * Calculate the distance between the current location and the given activity location.
+   *
+   * @param activityLocation The location of the activity.
+   * @return The distance in kilometers.
+   */
+  fun getDistanceFromCurrentLocation(activityLocation: Location?): Float? {
```
hmaryam84, line 70: Okay, I understand now your null handling. So you are returning null if one of the two input locations is null. I guess this is useful in the code, as the user may not disclose their location, or as some activities do not have a location. In this case your code will return null, which you should correctly manage in the UI.

app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
```diff
@@ -229,10 +233,23 @@ fun ActivityDetailsScreen(
...
-                        modifier = Modifier.testTag("locationText"))
+                    Column {
+                      Text(
+                          text = location?.name ?: "No location",
+                          modifier = Modifier.testTag("locationText"))
+                      if (distance != null) {
+                        val distanceString =
+                            "Distance : " +
+                                if (distance < 1) {
+                                  "${round(distance * 1000)}m"
```
hmaryam84, line 244: Great detail ! Indeed, for less than 1km distance, it is better to show the distance in meters, with a better precision. Great ! However, you did not implement tests that check if this works correctly...

app/src/main/java/com/android/sample/ui/listActivities/ViewListActivities.kt
```diff
@@ -84,6 +96,26 @@ fun ListActivitiesScreen(
...
+
+  LaunchedEffect(Unit) {
+    if (ContextCompat.checkSelfPermission(context, Manifest.permission.ACCESS_FINE_LOCATION) ==
+        PackageManager.PERMISSION_GRANTED) {
+      locationViewModel.fetchCurrentLocation()
+    } else {
+      locationPermissionLauncher.launch(Manifest.permission.ACCESS_FINE_LOCATION)
+    }
+  }
+
```
hmaryam84, line 118: This adds lots of code lines. I don't know if there is a more graceful way to check for this, as it is purely duplicated code from `ui/map/Map.kt` 
Also, this part is not tested (and is actually hard to test).
But it is great that we do not need to go through maps to load the distances.

I would recommend this permission grant is rather demanded when the app launches.

app/src/main/java/com/android/sample/ui/listActivities/ViewListActivities.kt
```diff
@@ -190,7 +223,8 @@ fun ActivityCard(
     navigationActions: NavigationActions,
     listActivitiesViewModel: ListActivitiesViewModel,
     profileViewModel: ProfileViewModel,
-    profile: User?
+    profile: User?,
+    distance: Float? = null
```
hmaryam84, line 227: Great you gave it a default value ! It helped you not change code in every place ActivityCard is used (such as the tests).

app/src/test/java/com/android/sample/model/map/LocationViewModelTest.kt
```diff
@@ -120,4 +121,34 @@ class LocationViewModelTest {
...
+        // Given
+        viewModel.setCurrentLocation(null)
+        val activityLocation = Location(1.0, 2.0, "Test Location")
+
+        // When
+        val distance = viewModel.getDistanceFromCurrentLocation(activityLocation)
+
+        // Then
+        assertEquals(null, distance)
+      }
```
hmaryam84, line 137: It is very important to check the robustness is case of null cases. Good.

app/src/test/java/com/android/sample/model/map/LocationViewModelTest.kt
```diff
@@ -120,4 +121,34 @@ class LocationViewModelTest {
...
+        viewModel.setCurrentLocation(currentLocation)
+        val activityLocation = Location(46.5375, 6.573611, "Epenex")
+        // When
+        val distance = viewModel.getDistanceFromCurrentLocation(activityLocation)
+        // Then
+        assertNotNull(distance)
+        if (distance != null) {
+          assertEquals(2.34f, distance, 0.05f)
+        }
+      }
```
hmaryam84, line 153: The delta is really important here ! Calculations may differ depending on formulas. Here you gave a 50m precision, which is adapted to our use case. Perfect!

hmaryam84 (approved):
I approve this PR. Thanks for your comprehensive description. The code is very clear to understand and well commented. I had a good time reviewing your work. On top of the distance being displayed to the users, you just built the foundation of our sorting recommendation algorithm, as you said. Thank you !

app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
```diff
@@ -289,6 +315,53 @@ class OverviewScreenTest {
...
+          profileViewModel = userProfileViewModel,
+          profile = testUser,
+          activity = activity,
+          distance = 0.5503f)
+    }
+    composeTestRule
+        .onNodeWithTag("distanceText", useUnmergedTree = true)
+        .assertTextContains("Distance : 550m")
+  }
+
```
hmaryam84, line 334: So Great that you test the UI now ! Plus you also test the rounding which is great.

PR #232, opened 2024-11-22
fix(resources): refactor activities in dummy data to match the new data class of interests.

aminebenji04:
In this PR I just handled problems due to the merge of the new interests data class


mohamedtahaguelzim (approved):
Not much to say. Indeed, using predefined interests makes much more sense than contextually writing them. However, we need to make sure that no other tests rely on their specific text. If the CI passes, then I'd approve it!

PR #237, opened 2024-11-22
Enhance Activity Scoring and Sorting Functionality

### LINKED ISSUE #228: implement an algorithm that sorts the activities in the overview according interests/categories and number of likes.

mohamedtahaguelzim:
## **Summary:**
This PR introduces significant improvements to the activity recommendation system by adding several scoring functions, caching mechanisms, and sorting features. Key functionalities were implemented to calculate activity scores based on various factors, such as distance, date, participation, completion ratio, price, and user interest. The system is now able to sort activities by their calculated scores, improving recommendations based on the user’s profile.

The formula used in the PR to calculate the activity score is based on a weighted sum of various factors, each contributing a specific weight to the final score. The formula is designed to evaluate an activity based on user preferences, such as distance, date, interest, participation, completion, and price. Here's a breakdown of the process:

### 1. **Weighted Criteria (Factors):**
The activity score is based on multiple criteria, each assigned a weight. These weights are defined in the `getWeights()` function:

```kotlin
return mapOf(
    "distance" to 0.2,
    "date" to 0.15,
    "interest" to 0.25,
    "participation" to 0.15,
    "completion" to 0.1,
    "price" to 0.15
)
```

- `distance`: 20%
- `date`: 15%
- `interest`: 25%
- `participation`: 15%
- `completion`: 10%
- `price`: 15%

### 2. **Score Calculation for Each Criterion:**
For each of the criteria, a function calculates a score based on the provided activity data:
- **Distance**: The `calculateDistanceScore` function computes the score based on the distance between the user and the activity. It returns a score that decreases with distance but is capped at 1 (max distance of 100 km).
- **Date**: The `calculateDateScore` function calculates a score based on how close the activity's date is to the current time. The score decreases as the time difference increases, up to a maximum of 96 hours.
- **Interest**: The `calculateInterestScore` function currently returns a default value of 0.0, as it's planned for future implementation in Sprint 7.
- **Participation**: The `calculateParticipationScore` function calculates the score based on the user's overlap in activities with the activity creator. It looks for matching activities from the creator and divides by 10 (to normalize the score).
- **Completion**: The `calculateCompletionScore` function calculates a score based on how full the activity is. It returns a score based on the ratio of participants to available spots, capped at 1.0.
- **Price**: The `calculatePriceScore` function calculates a score based on the price of the activity. It normalizes the score such that higher prices lower the score, capped at 1 for prices up to 100 CHF.

### 3. **Final Score Calculation:**
After calculating the individual scores for each factor, the scores are weighted according to their respective weights from `getWeights()`. The weighted sum is then normalized by dividing it by the total weight (which is the sum of the weights, i.e., 1.0 in this case).

```kotlin
val score =
    (distanceScore * weights["distance"]!! +
        dateScore * weights["date"]!! +
        interestScore * weights["interest"]!! +
        participationScore * weights["participation"]!! +
        completionScore * weights["completion"]!! +
        priceScore * weights["price"]!!) / totalWeights
```

This weighted sum represents the activity's score. If the score has been previously calculated for this activity (stored in `cachedScores_`), it returns the cached score for better performance.

### 4. **Caching:**
Once the score is calculated, it's stored in the `cachedScores_` map for later use. This ensures that repeated calculations for the same activity do not have to recompute the scores, improving performance.

```kotlin
cachedScores_[activity.uid] = score
```

### 5. **Final Output:**
The final activity score is then returned, reflecting the relative suitability of the activity based on the user's preferences and the activity's characteristics.

This formula provides a flexible and extensible way to evaluate and rank activities based on dynamic and customizable criteria. Each factor contributes to the overall score based on how important it is for the user (as defined by the weights).

---

### Testing :

This pull request includes unit tests for the `ScoreCalculator` in the `ListActivitiesViewModel`. The tests cover key functionalities:

1. **Weights Calculation:** Ensures the `getWeights` method returns correct weights for scoring factors.

2. **Distance Score:** Tests the `calculateDistanceScore` method for edge cases (e.g., `null`, max distance, valid distance).

3. **Date Score:** Verifies that `calculateDateScore` works for past, current, and future dates.

4. **Time Difference:** Ensures `calculateHoursBetween` correctly calculates time differences and handles invalid cases.

5. **Interest and Participation Scores:** Validates `calculateInterestScore` and `calculateParticipationScore` for edge cases (e.g., `null` lists).

6. **Completion Score:** Tests `calculateCompletionScore` for various participation scenarios (e.g., exceeding, equal, or fewer than spots).

7. **Price Score:** Verifies `calculatePriceScore` handles prices correctly, returning scores between `0.0` and `1.0`.

8. **Activity Sorting:** Tests `sortActivitiesByScore` for correct sorting by scores and handling empty lists.

9. **Activity Score Calculation:** Ensures `calculateActivityScore` computes scores correctly using weighted factors.

10. **Caching:** Verifies caching works for repeated score calculations.

11. **Null Distance:** Ensures the method handles `null` distance values without errors.

---

Closes #228

### DIFF
--- a/app/src/androidTest/java/com/android/sample/ui/activity/LikedActivitiesTest.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/activity/LikedActivitiesTest.kt
@@ -42,7 +42,7 @@ class LikedActivitiesTest {
     activitiesRepository = mock(ActivitiesRepository::class.java)
     profilesRepository = mock(ProfilesRepository::class.java)
     navigationActions = mock(NavigationActions::class.java)
-    viewModel = ListActivitiesViewModel(activitiesRepository)
+    viewModel = ListActivitiesViewModel(profilesRepository, activitiesRepository)
 
     // val userStateFlow = MutableStateFlow(testUser)
     navigationActions = mock(NavigationActions::class.java)
--- a/app/src/androidTest/java/com/android/sample/ui/activityDetails/ActivityDetailsScreenTest1.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/activityDetails/ActivityDetailsScreenTest1.kt
@@ -313,7 +313,8 @@ class ActivityDetailsScreenAndroidTest {
   @Test
   fun leaveActivityToast_displays_whenLeftActivity() {
     val mockActivitiesRepo = MockActivitiesRepository()
-    mockViewModel = spy(ListActivitiesViewModel(mockActivitiesRepo))
+    mockViewModel =
+        spy(ListActivitiesViewModel(mock(ProfilesRepository::class.java), mockActivitiesRepo))
 
     val mockProfileRepo = MockProfilesRepository()
     mockProfileViewModel = spy(ProfileViewModel(mockProfileRepo))
--- a/app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/listActivities/ListActivitiesTest.kt
@@ -99,7 +99,7 @@ class OverviewScreenTest {
     profilesRepository = mock(ProfilesRepository::class.java)
     activitiesRepository = mock(ActivitiesRepository::class.java)
     navigationActions = mock(NavigationActions::class.java)
-    listActivitiesViewModel = ListActivitiesViewModel(activitiesRepository)
+    listActivitiesViewModel = ListActivitiesViewModel(profilesRepository, activitiesRepository)
 
     testUser =
         User(
--- a/app/src/androidTest/java/com/android/sample/ui/map/MapScreenTest.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/map/MapScreenTest.kt
@@ -13,6 +13,7 @@ import com.android.sample.model.activity.ListActivitiesViewModel
 import com.android.sample.model.map.LocationPermissionChecker
 import com.android.sample.model.map.LocationRepository
 import com.android.sample.model.map.LocationViewModel
+import com.android.sample.model.profile.ProfilesRepository
 import com.android.sample.resources.dummydata.activity
 import com.android.sample.ui.navigation.NavigationActions
 import com.android.sample.ui.navigation.Screen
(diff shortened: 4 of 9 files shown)


Yassyass181104 (approved):
For me, everything looks good, the tests are complete and the code is implementing a good strategy to have a better display of the activities. Great work !

aminebenji04 (changes requested):
Your PR is very detailed and doesn't let me any questionning, thank you for the clarity. However, I am not sure that the caching technique you use is very good, what if an activity has 0 participants, and 1 hour after it has 8/10 participants, the caching will make this activity be displayed in the lower parts because the cached value of completion is 0 and not ten.
Is it something you thought about and do you have solutions? 
A solution I suggets is implementing a TTL for caches, so that when this TTL expires, the scores are recomputed, and it would also allow scores to be computed independently and not at the same time for all activities.

mohamedtahaguelzim:
> Your PR is very detailed and doesn't let me any questionning, thank you for the clarity. However, I am not sure that the caching technique you use is very good, what if an activity has 0 participants, and 1 hour after it has 8/10 participants, the caching will make this activity be displayed in the lower parts because the cached value of completion is 0 and not ten. Is it something you thought about and do you have solutions? A solution I suggets is implementing a TTL for caches, so that when this TTL expires, the scores are recomputed, and it would also allow scores to be computed independently and not at the same time for all activities.

Hi @aminebenji04,
Thanks for pointing this out. However, I disagree with your suggestion for the two following reasons : 

1. The current caching has an implicit "TTL", which is that it gets cleaned and recalculated every time the app is launched. This means that if the user closes the app, and starts it back, the recommendation score will be recalculated. So, since we do not expect the user to be in the app for 1 full hour, then, if they open the app an hour later, the score will indeed be recalculated. We shall note that the caching only affects the ranking, and if the activity data changes (e.g. new participants, new title, ...) this will be correctly synced, even without quiting the app.
2. What you are recommending is to use a TTL. So if I have 5 minutes caching, then 5 minutes later the activities will be reorganized according to their new score. I have considered this before starting to code, and I have concluded that this is NOT a good practice. When a user looks at the activities, they expect them to stay where they are, while using the app. Do you imagine using instagram, and then, randomly, the post your are looking at in your feed, disappears because the algorithm recalculated its recommendation score ?

These are my considerations, please tell me if you disagree with any of my points.
Thanks.

aminebenji04:
Ok super, if this is already implemented, there is nothing more to say!

aminebenji04 (approved):

PR #243, opened 2024-11-26
Disable button/database interactions when user is not connected to the network

aminebenji04:
In this PR, I tried to find all the buttons that may interact with the database to disable them using the previously created function performOfflineAwareAction so that there is no uncoherency between what we see in the screen and what happens in the databse, for example:

- The "Save Changes" button in the Profile Screen is now disabled when the phone is offline. Previously, users could click the button, leading to no updates in the database due to the lack of network connectivity, causing confusion.
- Buttons to enroll or leave activities in the Overview or Activity Details screen are wrapped with performOfflineAwareAction. When offline, the buttons are disabled and provide feedback via a toast, preventing users from attempting actions that won't sync with the database.

**How to Test**
- Run the app and navigate to the affected screens.
- Simulate offline mode by disabling the device's network connection.
- Validate that:
   - Database interactions are blocked with appropriate feedback.
   - Navigation to restricted screens is prevented with user notifications.
   - Existing functionality (e.g., data viewing) remains unaffected in offline mode.

### DIFF
--- a/app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
+++ b/app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
@@ -401,7 +401,12 @@ fun ActivityDetailsScreen(
                 } else {
                   // Creator of the activity
                   Button(
-                      onClick = { navigationActions.navigateTo(Screen.EDIT_ACTIVITY) },
+                      onClick = {
+                        performOfflineAwareAction(
+                            context = context,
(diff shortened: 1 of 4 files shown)


Yassyass181104 (approved):
I've seen the whole changes. It's very good, you keep the same implementation but ensure that the necessary actions can be done only if connected to internet. You do the verification through performOfflineAwareAction, isn't it ? The code is clean and there is no point by going file by file since the logic stays the same. Very good !

PR #268, opened 2024-11-28
Feature/activity to map

### LINKED ISSUE #266: Add a button to go to the map
- [x] Include a button in the activity details screen to navigate directly to the map view.
- [x] When clicked, the map should be centered around the related activity pin

### LINKED ISSUE #267: Map automatically centered around your localisation
- [x] If localisation is active, then automatically center the map around the current localisation.

elvan03:
In this PR, I solve 2 issues. I decided to do them in the same branch since it was faster for me and the second issue was a small task.
1) I allow to go directly from the activity detail view to the map view, centered around the pin of the activity which detail was opened. In the activity detail, I decided to use a string "See on map" rather than a button, as it's better for UI, at least right now.
2) By default, I center the map view around the user's position.

As you can see on my commits, I initially decided to change the navigation and to add an optional location parameter to the map screen and navigateTo methods, in order to give the location while navigating to the map (and center around the location I gave). This was really hard to implement because it meant changing a lot of things in navigation and adding a lot of logic in the main, when setting navHost. 

Eventually, I was suggested by scrum master to use an existing parameter in the map screen, selectedActivity, which gives the latest opened activity detail. 
That's what I did, along with a check that user is indeed coming from the detail screen and not the bottom bar. In the case we come from bottom bar, the default location displayed is user's location.

closes #266 
closes #267


app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
```diff
@@ -260,6 +263,16 @@ fun ActivityDetailsScreen(
...
+                        Spacer(modifier = Modifier.height(STANDARD_PADDING.dp))
+                        // text field button to navigate to the activity's location on the map
+                        // screen
+                        Text(
+                            text = stringResource(id = R.string.button_to_map),
+                            modifier =
+                                Modifier.testTag("activityToMapText")
+                                    .clickable(
+                                        onClick = { navigationActions.navigateTo(Screen.MAP) }),
+                            style = TextStyle(textDecoration = TextDecoration.Underline))
```
hmaryam84, line 275: Good for the string that was added in the string.xml file and also for having chosen a good text style.

app/src/main/java/com/android/sample/ui/map/Map.kt
```diff
@@ -47,36 +63,45 @@ import com.google.android.gms.maps.CameraUpdateFactory
...
       try {
-        val loc =
-            (activities as ListActivitiesViewModel.ActivitiesUiState.Success)
-                .activities
-                .firstNotNullOf { it.location }
-        LatLng(loc.latitude, loc.longitude)
+        // the first location that displays is the last activity detail that was checked
+        // if you don't come from an activity detail, it will be centered around user's position
+        if (previousScreen == Screen.ACTIVITY_DETAILS) {
+          activityDetail?.location?.let { LatLng(it.latitude, it.longitude) } ?: defaultLocation
```
hmaryam84, line 95: The best for this would have been a view model, but since our case is simple we can go with this implementation. Note that this works only because we don’t have a bottom bar in the details screen.

app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
hmaryam84: this test Tag haven't been tested.

hmaryam84:
The features seem correctly implemented.
The best for the button that navigates from the activity details to the map would have been a view model, but since our case is simple we can go with this implementation. Note that this works only because we don’t have a bottom bar in the details screen. There are no tests and There is one test tag that haven't been tested. This could be improved later even if the coverage would not change that much. I will merge because the coverage on new code is more than 80% and because the functionnality seems to work well on my emulator .

hmaryam84 (approved):
The features seem correctly implemented.
The best for the button that navigates from the activity details to the map would have been a view model, but since our case is simple we can go with this implementation. Note that this works only because we don’t have a bottom bar in the details screen. There are no tests and There is one test tag that haven't been tested. This could be improved later even if the coverage would not change that much. I will merge because the coverage on new code is more than 80% and because the functionnality seems to work well on my emulator .

app/src/main/java/com/android/sample/ui/map/Map.kt
elvan03: Yes, this was discussed and approved with the scrum master before. PS: what do you mean with a view model ?

app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
elvan03: Yes, I have warned you privately on this. Because of some technical issues and because I had not much time left (a lot of merges waiting on pipeline) I decided to leave this for next week.

PR #277, opened 2024-12-02
Fix bug when start time is after ending time

Yassyass181104:
This pull request includes changes to ensure that the validation logic for activity start and end times is correctly implemented in the `CreateActivityScreen` and `EditActivityScreen` components. The main changes involve correcting the condition used to check if the start time is before the end time.

Changes to validation logic:

* [`app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt`](diffhunk://#diff-38c302af30fc7257980234ed59aed91589cb302671f5734b59e740044b451e8aL528-R528): Updated the condition in `CreateActivityScreen` to correctly check if the start time is before the end time.
* [`app/src/main/java/com/android/sample/ui/activity/EditActivity.kt`](diffhunk://#diff-cea2a1072c58261b4a239e412749fe81981de842d5e34d63e75f8bec747c3144L529-R529): Updated the condition in `EditActivityScreen` to correctly check if the start time is before the end time.


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -539,7 +539,7 @@ fun CreateActivityScreen(
                               "You must be logged in to create an activity.",
                               Toast.LENGTH_SHORT)
                           .show()
-                    } else if (hourDateViewModel.isBeginGreaterThanEnd(startTime, duration)) {
```
DAFERMOH, line 542: checks if start time is before the end time. (if not an error toast will be displayed in the screen with the message: "The start time must be before the end time.".

app/src/main/java/com/android/sample/ui/activity/EditActivity.kt
```diff
@@ -526,7 +526,7 @@ fun EditActivityScreen(
             Button(
                 enabled = title.isNotEmpty() && description.isNotEmpty(),
                 onClick = {
-                  if (hourDateViewModel.isBeginGreaterThanEnd(
```
DAFERMOH, line 529: checks if start time is before the end time. (if not an error toast will be displayed in the screen with the message: "The start time must be before the end time.".

DAFERMOH (approved):
These changes correct the logic to make sure the start time is before the end time, which aligns with the intent of the toast messages indicating time validation. 
Good job fixing this bug!

PR #282, opened 2024-12-02
Add caching for the profile in profile screen

aminebenji04:
In this pull request, I added the same logic as in liked activities but to display the past activities, the enrolled activities and the created activities stored in the local database when the user is offline. 
### How to try the new feature:

1. Disable Wi-Fi and data
2. Open the App
3. Go to the profile screen
4. You should see all the activities in your profile screen even if you have no connection

### DIFF
--- a/app/src/test/java/com/android/sample/model/profile/ProfileViewModelTest.kt
+++ b/app/src/test/java/com/android/sample/model/profile/ProfileViewModelTest.kt
@@ -1,11 +1,15 @@
 package com.android.sample.model.profile
 
 import androidx.test.core.app.ApplicationProvider
+import com.android.sample.model.activity.database.AppDatabase
+import com.android.sample.model.profile.database.UserDao
 import com.android.sample.resources.dummydata.activity
 import com.android.sample.resources.dummydata.testUser
 import com.google.firebase.FirebaseApp
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.firestore.FirebaseFirestore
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/android/sample/model/profile/ProfileViewModelTest.kt
```diff
@@ -192,4 +195,33 @@ class ProfileViewModelTest {
     profileViewModel.clearUserData()
     assertEquals(null, profileViewModel.userState.value)
   }
+
+  @Test
+  fun fetchUserDataCachesLocallyOnSuccess() = runTest {
+    val mockUser = testUser // Replace with your test user object
```
Yassyass181104, line 201: Passing a testuser from dummyData into a local variable in your test isn't really usefull

app/src/test/java/com/android/sample/model/profile/ProfileViewModelTest.kt
Yassyass181104, line 1: This new test verifies that the existence or non existence of the profile is well handled, well done

Yassyass181104 (changes requested):
These new lines of code are perfectionements that are very usefull. You may have to change your new test by using directly testuser from dummyData but the rest is very good. Nice !

app/src/test/java/com/android/sample/model/profile/ProfileViewModelTest.kt
Yassyass181104, line 1: Cleaner since the changes, very good

Yassyass181104 (approved):

PR #286, opened 2024-12-04
Make pro creator not automatically in the list of participants

Yassyass181104:
This pull request includes changes to the `CreateActivity.kt` file to enhance functionality and improve user experience. The most important changes include the addition of a new import for `ActivityType` and the inclusion of a conditional check for `selectedOptionType` when adding attendees.

Enhancements to functionality:

* [`CreateActivity.kt`](diffhunk://#diff-38c302af30fc7257980234ed59aed91589cb302671f5734b59e740044b451e8aR60): Added import for `ActivityType` to support new activity type functionality.

Improvement to user experience:

* [`CreateActivity.kt`](diffhunk://#diff-38c302af30fc7257980234ed59aed91589cb302671f5734b59e740044b451e8aR557): Added a conditional check to ensure attendees are only added if the `selectedOptionType` is `ActivityType.INDIVIDUAL.`.

### DIFF
--- a/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
+++ b/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
@@ -555,7 +555,7 @@ fun CreateActivityScreen(
                           .show()
                     } else {
                       if (selectedOptionType == ActivityType.INDIVIDUAL.name)
-                          attendees += profileViewModel.userState.value!!
+                          profileViewModel.userState.value?.let { user -> attendees += user }
                       try {
                         imageViewModel.uploadActivityImages(
                             activityId,
--- a/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
+++ b/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
@@ -57,6 +57,7 @@ import androidx.compose.ui.unit.sp
 import com.android.sample.R
 import com.android.sample.model.activity.Activity
 import com.android.sample.model.activity.ActivityStatus
+import com.android.sample.model.activity.ActivityType
 import com.android.sample.model.activity.ListActivitiesViewModel
 import com.android.sample.model.activity.categories
 import com.android.sample.model.activity.types
@@ -553,7 +554,8 @@ fun CreateActivityScreen(
                       Toast.makeText(context, "You must select a location.", Toast.LENGTH_SHORT)
                           .show()
                     } else {
-                      attendees += profileViewModel.userState.value!!
+                      if (selectedOptionType == ActivityType.INDIVIDUAL.name)
+                          attendees += profileViewModel.userState.value!!
                       try {
                         imageViewModel.uploadActivityImages(
                             activityId,


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -553,7 +554,8 @@ fun CreateActivityScreen(
                       Toast.makeText(context, "You must select a location.", Toast.LENGTH_SHORT)
                           .show()
                     } else {
-                      attendees += profileViewModel.userState.value!!
+                      if (selectedOptionType == ActivityType.INDIVIDUAL.name)
```
hmaryam84, line 557: The !! are known to be unsafe, I suggest modifu^ying it to something like: 
profileViewModel.userState.value?.let { attendees += it } 
or attendees += profileViewModel.userState.value ?: emptyList()

hmaryam84 (changes requested):
Even if the modification seems small, it could be source of problems in the future, Please try to remove the unsafe !! and replace it with one of the suggested lines of code.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
mohamedtahaguelzim: > I think you miss the fact that the '!!' are used to say : "I'm sure this is never null". Knowing that this screen is allowed to access iff we are logged in (i.e. the user is not null) then even if it's unsafe here we can use it correctly

Saying "this screen is allowed access iff we are logged in" is not enough. Because we never know, say in the future we decide to allow guests to create activities. I believe what you say could be correct if we have a check **within** the screen, that would display an error message such as "You are not logged in, you can not create an activity".
I guess one could argue that this check is done thanks to the `if (creator == "") ... else ...` but this check is, in my opinion, not explicit enough.

I agree with you @Yassyass181104 that, knowing how we implemented the things, this should not pose any issue. But I also agree with @hmaryam84 on the fact that it is a bad practice, and that such behaviour already caused us crashes, e.g. the crash of the app when we opened some activities in the last meeting with the coaches.

More generally, I'd advise you to never assume anything (e.g. "I'm sure this is never null"), and to always consider that the worst case could happen!

@hmaryam84 suggested : 
> profileViewModel.userState.value?.let { attendees += it }
or attendees += profileViewModel.userState.value ?: emptyList()
I'd approve these suggestions.
For you to implement them quicker, I'll do you a suggestion directly on GitHub, and you can accept it here directly.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -553,7 +554,8 @@ fun CreateActivityScreen(
...
+                          attendees += profileViewModel.userState.value!!
```
mohamedtahaguelzim, line 558: Credit to @hmaryam84 for this suggestion :

```suggestion
                          attendees += profileViewModel.userState.value ?: emptyList()
```

mohamedtahaguelzim: You can directly accept the suggestion from here, and do a commit from GitHub.

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
Yassyass181104: Didn't see this point of view, even thought norammly we can't be able to create activities being a guest but I totally agree with both of you for the fact that it's a bad practice

Yassyass181104:
> The feature seems now correctly implemented and safe after the change tht was made. There are no tests for it, but since it is only one line of code, this is not a big problem.

PR #295, opened 2024-12-04
feat(map): replace the old blue marker by a blue circle for the current location

aminebenji04:


app/src/main/java/com/android/sample/ui/map/Map.kt
Yassyass181104, line 1: Good change, please avoid magic values

Yassyass181104 (changes requested):
Good but uses magic values

Yassyass181104 (approved):
Even better now !

PR #297, opened 2024-12-04
Fix/nominatim suggestions unfocus keyboard

### LINKED ISSUE #164: Nominatim Suggestions unfocus keyboard
When writing a location (Activity Creation Screen), the nominatim suggestions panel unfocuses the keyboard, leading to a very bad ux.

I have had this issue in the bootcamp, and I fixed it. I will apply the same fix for this case.

mohamedtahaguelzim:
### **Pull Request Overview**

This pull request enhances the location suggestion feature in `CreateActivity.kt` and `EditActivity.kt` by implementing a custom dropdown menu using `Box` and `DropdownMenu`. The changes improve user experience by keeping the text field active during interactions and ensuring the dropdown dynamically updates with suggestions.

### **Key Changes**

1. **Custom Dropdown Implementation**  
   - Replaced `ExposedDropdownMenu` with a `DropdownMenu` inside a `Box` layout for better flexibility and control.  
   - Used `PopupProperties(focusable = false)` to retain text field focus during dropdown interactions.

2. **Dynamic Suggestions Handling**  
   - Dropdown visibility updates based on user input, showing relevant suggestions when the text field is not empty.  
   - Displays up to three suggestions with a "More..." option for extended functionality.

3. **Improved UI Feedback**  
   - Dropdown behavior is smooth and intuitive, avoiding interruptions caused by focus shifts.
   
### **Note on duplication :**
This PR is intended to fix the bug we were encountering, not necessarily optimize code implementation. While I am aware that this results in a highly duplicated code in my PR, I consider optimizing it as out of scope, since the original code itself is highly duplicated, and should be reconsidered by its original author @Yassyass181104.

Closes #164


aminebenji04 (approved):
This is a good PR, this problem was very disturbing, I don't think tests are needed as it doesn't bug on testing but when interacting with the location text field with the emulator it bugs a lot, I tried it in your branch and everything is fine. Thank you !

PR #322, opened 2024-12-05
Refactor/bottom bar

GhaliBerbich:
Just Removed the text from the bottom bar buttons.


aminebenji04 (approved):
Very good, nothing to add here !

PR #346, opened 2024-12-09
fix: remove duplicated code

### LINKED ISSUE #345: bug: Location selection is not possible in create screen
Location selection in the create activity screen is not possible due to some location text being superposed to the text field. This is probably due to some duplicated code.

Will fix it myself ASAP.

mohamedtahaguelzim:
Very simple PR with a codebase simplification:
* Removed redundant `DropdownMenuItem` creation for location suggestions.
* Replaced hardcoded `30` with `TOP_TITLE_SIZE` constant for limiting location name length.

This was causing a UI bug in the location selection process, which stopped us from correctly creating an activity.

Closes #345


elvankonukseven (approved):
Thanks for solving this problem! LGTM!

PR #359, opened 2024-12-11
test/notifications_24h_before

### LINKED ISSUE #356: add tests and refactor notification setup code

elvan03:
In this PR, I added some tests for the previous feature I added, notifications 24h before start time of activity. Due to the nature of notification helper and notification receiver files, I added both android and unit tests for these files.

closes #356


app/src/androidTest/java/com/android/sample/helper/NotificationHelperInstrumentedTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    // Verify the scheduled PendingIntent
+    val expectedRequestCode = activityId.toIntOrNull() ?: activityId.hashCode()
+    val pendingIntent =
+        PendingIntent.getBroadcast(
+            context,
+            expectedRequestCode,
+            Intent(context, NotificationReceiver::class.java),
+            PendingIntent.FLAG_NO_CREATE or PendingIntent.FLAG_IMMUTABLE)
+    assertTrue(pendingIntent != null)
+  }
```
aminebenji04, line 64: Very good, you test that the pending intent is not null. It looks like it works fine !

app/src/androidTest/java/com/android/sample/receivers/NotificationReceiverAndroidTest.kt
```diff
@@ -0,0 +1,92 @@
...
+  fun testNotificationNotCreatedWithMissingTitle() {
+    val intent =
+        Intent().apply {
+          putExtra("activityId", "123")
+          putExtra("activityName", "Test Activity")
+          // Missing title
+        }
+    receiver.onReceive(context, intent)
+    verify(notificationManager, never()).notify(any(), any())
+  }
```
aminebenji04, line 67: Good but when would this happen? When would an intent miss the title or the name of an activity ?

app/src/test/java/com/android/sample/receivers/NotificationReceiverUnitTest.kt
```diff
@@ -0,0 +1,82 @@
...
+    val notification = notifications[0]
+
+    val shadowNotification = Shadows.shadowOf(notification)
+    assert(shadowNotification.contentTitle == notificationTitle) {
+      "Notification title does not match"
+    }
+    assert(shadowNotification.contentText == "Reminder: $activityName") {
+      "Notification content does not match"
+    }
+  }
```
aminebenji04, line 59: Very good, this was the test case I was looking for ! So if and only if the intent has all the info needed, it can send the notification, good job !

app/src/androidTest/java/com/android/sample/receivers/NotificationReceiverAndroidTest.kt
elvan03: In notification receiver class, I implemented an early return if required notification data (title, activity name) is missing, preventing incomplete notification. This is an additional test to verify the same thing.

aminebenji04 (changes requested):
There are a lot of test duplicated for the tests of NotificationReceiver in android tests and unit tests

elvan03:
Thanks for your remark! Indeed, I didn't realise I was testing the same thing twice. 
It should be good now :)

app/src/test/java/com/android/sample/model/receivers/NotificationReceiverTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.kotlin.any
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.never
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+@RunWith(AndroidJUnit4::class)
+class NotificationReceiverTest {
```
aminebenji04, line 18: All the "not working" tests cases are tested very good! And the case where the notification is created is in the android test, these can be unified but it is not a big matter

aminebenji04 (approved):
Very good PR, this should significantly increase the coverage and this ensures that the notifications don't have buggy cases, LGTM! Also the requested changes were handled in a good way !

PR #361, opened 2024-12-11
Feature/notification-deletedActivities

### LINKED ISSUE #358: if the creator deletes the activity, attendees get a notif and creator gets punished on his rating , if its late minute.
- [ ] Send a notification to users when the creator deletes an activity they were enrolled in. 
- [ ] If the deletion occurs at the last minute, reduce the creator’s rating as a penalty. 
- [ ] Define "last-minute" based on a specific timeframe, and ensure the penalty system integrates with the rating mechanism.

elvan03:
In this PR, I created notifications that are sent to the participants when an activity they are enrolled to is deleted. 

When the activity is a "PRO", the creator does not receive a notification as he's not a participant, when the activity is of type "INDIVIDUAL", then all participants, including the creator, receive notifications. 

The tests are also included in this PR + everything works fine on the emulator.

Here is a visual of the notification:

Screenshot 2024-12-12 at 22 10 50

The notification string can easily be changed if needed :)

closes #358


app/src/test/java/com/android/sample/model/receivers/NotificationReceiverTest.kt
```diff
@@ -1,92 +0,0 @@
-package com.android.sample.receivers
```
elvan03, line 1: I deleted this file because I noticed I made a mistake by duplicating it. This exact file is also in the android test for notification receiver tests, where it's supposed to be.

app/src/test/java/com/android/sample/receivers/NotificationReceiverUnitTest.kt
```diff
@@ -57,4 +58,39 @@ class NotificationReceiverUnitTest {
...
+      "Notification title does not match"
+    }
+    assert(shadowNotification.contentText == "Join or create another activity!") {
+      "Deletion notification content does not match"
+    }
+    assert(notification.category == NotificationCompat.CATEGORY_EVENT) {
+      "Notification category should be EVENT for deletion notifications"
+    }
+  }
 }
```
GhaliBerbich, line 96: Good job for your current tests for the helper and receiver and the 100% coberage achieved on them. There are no current tests for ScheduleNotification and sendDeletionNotification that are in the app file, you could maybe add test for them or for the code you added in the listActivityViewModel when an activity is deleted to reach 80% of coverage on new code.  Also you should use users from dummy data rather than creating new users each time for your list of participants.

GhaliBerbich (changes requested):

app/src/androidTest/java/com/android/sample/helper/NotificationHelperInstrumentedTest.kt
```diff
@@ -62,4 +66,107 @@ class NotificationHelperInstrumentedTest {
...
+    val alarmManager = context.getSystemService(Context.ALARM_SERVICE) as AlarmManager
+    val newTime = System.currentTimeMillis() + TimeUnit.MINUTES.toMillis(1)
+
+    try {
+      alarmManager.setExactAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, newTime, newPendingIntent)
+      assertTrue(true) // If we get here, the previous alarm was successfully cancelled
+    } catch (e: Exception) {
+      fail("Failed to set new alarm, previous one might not be properly cancelled: ${e.message}")
+    }
+  }
```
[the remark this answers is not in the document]
elvan03: Thanks for the feedback! Actually, I think the code uses Android-specific components that require a running Android environment here, like AlarmManager or PendingIntent. I tried a lot to put the tests in the same file but I really think in this case it's hard and better to separate tests in 2 classes, android and unit

app/src/test/java/com/android/sample/receivers/NotificationReceiverUnitTest.kt
elvan03: In think the test for App.kt has to be handled with the end-to-end tests as it contains app initialization. Concerning the other tests, we will be refactoring all tests before deadline always. Right now I have a technical issue and unfortunately can't add a test in listActivityViewModel tests

GhaliBerbich: ok i see, this is okay then

PR #363, opened 2024-12-11
Refactor Profile Creation Screen to Match Figma Designs

### LINKED ISSUE #338: Improve the profile creation screen based on the figma

aminebenji04:
Fixes #338
In this PR, none of the business logic was touched, but IO refactored a lot of UI:
- I added colors for interest categories
- I displayed colors for each interest in profile creation
- I adjusted the UI of the Profile reation Screen to display name and surname in one row
- I added textures to all the textfields thanks to the card
### - Here is the UI we should all follow for our ui elements (text fields, buttons etc.):
```kotlin
            Card(
                modifier = Modifier.fillMaxWidth(WIDTH_FRACTION_SM).testTag("PurposeCard"),
                colors =
                    CardDefaults.cardColors(containerColor = MaterialTheme.colorScheme.background),
                elevation =
                    CardDefaults.cardElevation(defaultElevation = CARD_ELEVATION_DEFAULT.dp),
                shape = RoundedCornerShape(ROUNDED_CORNER_SHAPE_DEFAULT.dp),
            ) {
                  OutlinedButton() // or textField etc...
                  }
``` 
Here you can see the changes added: 
image

### DIFF
--- a/app/src/main/java/com/android/sample/model/profile/User.kt
+++ b/app/src/main/java/com/android/sample/model/profile/User.kt
@@ -1,7 +1,14 @@
 package com.android.sample.model.profile
 
+import androidx.compose.ui.graphics.Color
 import androidx.room.Entity
 import androidx.room.PrimaryKey
+import com.android.sample.resources.C.Tag.ART_ACTIVITY_COLOR
+import com.android.sample.resources.C.Tag.CULTURAL_ACTIVITY_COLOR
+import com.android.sample.resources.C.Tag.INDOOR_COLOR
+import com.android.sample.resources.C.Tag.MUSICAL_ACTIVITY_COLOR
+import com.android.sample.resources.C.Tag.OUTDOOR_COLOR
+import com.android.sample.resources.C.Tag.SPORT_COLOR
 
 @Entity(tableName = "User")
 data class User(
@@ -24,3 +31,12 @@ val InterestCategories =
         "Cultural Activity",
         "Art Activity",
         "Musical Activity") + "None"
+val InterestCategoriesColors =
+    mapOf(
+        "Sport" to SPORT_COLOR,
+        "Outdoor Activity" to OUTDOOR_COLOR,
+        "Indoor Activity" to INDOOR_COLOR,
+        "Cultural Activity" to CULTURAL_ACTIVITY_COLOR,
+        "Art Activity" to ART_ACTIVITY_COLOR,
+        "Musical Activity" to MUSICAL_ACTIVITY_COLOR,
+        "None" to Color.Gray)
--- a/app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
+++ b/app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
@@ -4,7 +4,10 @@ import android.graphics.Bitmap
 import android.util.Log
 import androidx.camera.view.CameraController
 import androidx.camera.view.LifecycleCameraController
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
```diff
@@ -134,88 +148,116 @@ fun ProfileCreationScreen(
...
+                    onValueChange = { surname = it },
+                    label = "Surname",
+                    validation = { input ->
+                      if (input.isBlank()) "Surname cannot be empty" else null
+                    },
+                    externalError = surnameErrorState.value,
+                    modifier = Modifier.weight(1f),
+                    errorTestTag = "surnameError",
+                    testTag = "surnameTextField")
+              }
```
GhaliBerbich, line 199: I think that it could be possible to put this code in a method that would be called here and that would only need the value of the name and surname, making the code of this screen clearer. It is already functionnal and good, but this could enhance code readability

app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
```diff
@@ -134,88 +148,116 @@ fun ProfileCreationScreen(
...
+              modifier =
+                  Modifier.testTag("uploadPicture")
+                      .clickable { showDialogImage = true } // Handle click action
+                      .padding(MEDIUM_PADDING.dp)
+                      .background(Color.Transparent)) {
+                Icon(
+                    imageVector = Icons.Default.AddAPhoto,
+                    contentDescription = "Add a photo",
+                    tint = Color.Black)
               }
```
GhaliBerbich, line 172: I think that this part can be modularized in a mthod so it will also possible to call it from editProfile for example

app/src/main/java/com/android/sample/model/profile/User.kt
```diff
@@ -24,3 +31,12 @@ val InterestCategories =
...
         "Musical Activity") + "None"
+val InterestCategoriesColors =
+    mapOf(
+        "Sport" to SPORT_COLOR,
+        "Outdoor Activity" to OUTDOOR_COLOR,
+        "Indoor Activity" to INDOOR_COLOR,
+        "Cultural Activity" to CULTURAL_ACTIVITY_COLOR,
+        "Art Activity" to ART_ACTIVITY_COLOR,
+        "Musical Activity" to MUSICAL_ACTIVITY_COLOR,
+        "None" to Color.Gray)
```
GhaliBerbich, line 42: These will be reduced to the categories used in activities in a next PR

GhaliBerbich (changes requested):

app/src/main/java/com/android/sample/model/profile/User.kt
aminebenji04: Yes exactly !

app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
aminebenji04: I understand the notion of abstracting this block into a reusable method for better readability. But given the context in which it is being used and the complexity of the arguments, it may not be the best approach here. Here's why:

- Dynamic Parameters Management: each TextFieldWithErrorState has a few parameters like value, onValueChange, validation, and test tags. Having these in a single method would either require more parameters or some sort of complex parameters setup that might be less explicit and harder to adjust at some later time.

- Limited Reusability: Although the abstraction makes this screen easier to understand, it can probably not be used in other parts of the app without major changes. This reduces the advantages of using such abstraction.

- Future Extensibility: The method should accommodate any addition of fields or rules in the future, which may make the design complex. Keeping the code clear here makes it easy to understand and change each field's logic.

GhaliBerbich (approved):
Thank you, now it is clearer.

PR #385, opened 2024-12-13
fix(time parse)

GhaliBerbich:
This PR concerns an unexpected nug. When logging into the apèp, an error in the listActivitiesViewModel hence in the overview arises from the HourViewModel where a parse triggers an error that is not caught and makes the app crash. 
- To fix that, I added a try on clause, and when an error is caught the date and time is set to the current, at the timezone of the device.

- The default value likely needs to be changed, but for now it is to make the app not crash.

### DIFF
--- a/app/src/main/java/com/android/sample/model/hour_date/HourDateViewModel.kt
+++ b/app/src/main/java/com/android/sample/model/hour_date/HourDateViewModel.kt
@@ -8,6 +8,7 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 import java.time.ZoneId
 import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeParseException
 
 open class HourDateViewModel : ViewModel() {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/android/sample/model/hour_date/HourDateViewModel.kt
```diff
@@ -49,17 +50,26 @@ open class HourDateViewModel : ViewModel() {
...
+      val combinedDateTime = LocalDateTime.of(localDate, localTime)
 
-    // Convert back to Firebase Timestamp
-    return Timestamp(
-        combinedDateTime.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli() / 1000, 0)
+      // Convert back to Firebase Timestamp
+      return Timestamp(
+          combinedDateTime.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli() / 1000, 0)
+    } catch (e: DateTimeParseException) {
+      // Log the error or handle it as needed
```
DAFERMOH, line 67: This is an acceptable solution as It does not crash the app but if we're getting the activity's TimeStamp we cannot give it the current time if It fails. We can instead throw exceptions if the parameters of the method " fun combineDateAndTime(date: Timestamp, time: String): Timestamp {" do not provide the correct parameters.
I have seen in that the app crash at: com.android.sample.ui.listActivities.ViewListActivitiesKt$ListActivitiesScreen$4.invoke(ViewListActivities.kt:178)
maybe we could instead check that the provided date is a Timestamp is not null and if the time is in the correct format: "HH:MM".

DAFERMOH (approved):
As I explained this is an acceptable and a temporary solution for a bug of such a priority (App crash when launching). Even though your solution could be improved I will aprove your changes but I will mention the improvement explained as an issue that needs to be done soon.

PR #389, opened 2024-12-13
Refactor the "no connection screen" 

aminebenji04:
In this PR, I refactored the "no connection screen" to display a fun animation of our logo spinning, and the connection message blinking in a fancy way. You can see more details here: 
image


app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
```diff
@@ -1,41 +1,87 @@
...
-            style = TextStyle(fontWeight = FontWeight.SemiBold, fontSize = SUBTITLE_FONTSIZE.sp))
+fun NoInternetScreen() {
+  val isSpinning = true
+  val spinningDuration = 1500
+  val fadingDuration = 1000
+  Box(
+      modifier = Modifier.fillMaxSize().background(Color.Black.copy(alpha = 0.5f)),
+      contentAlignment = Alignment.Center) {
+        val logo = painterResource(id = R.drawable.aptivity)
+        if (isSpinning) {
```
Yassyass181104, line 34: Since isSpinning is always true you can just replace it by true

app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
```diff
@@ -1,41 +1,87 @@
...
+
+          // Animation for rotation
+          val rotation by
+              infiniteTransition.animateFloat(
+                  initialValue = 0f,
+                  targetValue = 360f,
+                  animationSpec =
+                      infiniteRepeatable(
+                          animation =
+                              tween(durationMillis = spinningDuration, easing = LinearEasing)),
```
Yassyass181104, line 45: Would be good to replace spinningDuration by a constant placed in C

app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
```diff
@@ -1,41 +1,87 @@
...
+                  label = "")
+
+          // Animation for fading
+          val alpha by
+              infiniteTransition.animateFloat(
+                  initialValue = 0f,
+                  targetValue = 1f,
+                  animationSpec =
+                      infiniteRepeatable(
+                          animation = tween(durationMillis = fadingDuration, easing = LinearEasing),
```
Yassyass181104, line 55: Would be good to replace fadingDuration by a constant placed in C

Yassyass181104 (changes requested):
Overral very good but there are some points that need to be replaced just to have cleaner code but the implementation itself is very good and well done

app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
aminebenji04: It was done in purpose, I don't think that we will need this and adding this in constants would just make it more confusing

app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
aminebenji04: t was done in purpose, I don't think that we will need this and adding this in constants would just make it more confusing

app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
aminebenji04: t was done in purpose, I don't think that we will need this and adding this in constants would just make it more confusing

app/src/main/java/com/android/sample/ui/components/NoInternetScreen.kt
Yassyass181104: To answer this comment and the remaining, the point of using constants is to avoid using numbers in the functional classes and use the ones “provided” in the C file. For the boolean just since it’s not reassigned you can just either just delete the if condition or change it by true.

Yassyass181104 (approved):
You correctly corrected the code following the point of improvement pointed sooner during the review. All is looking good and clean, good job !

PR #390, opened 2024-12-13
Remove "more..." button in the location query

aminebenji04:
In this PR, I decided (after discussions with you) to remove the more... button as the first idea was to make it show 6 locations instead of 3, but it was very heavy for the app and made a lot of bugs, so removing it is not a big deal, because when being precise enough we can always find the location we are looking for


Yassyass181104 (approved):
Very good, this part of the implementation was useless and removing it is way better. Well done !

PR #393, opened 2024-12-14
Feature/auth/remove greetings screen

aminebenji04:


app/src/main/java/com/android/sample/ui/authentication/ChooseAccount.kt
Yassyass181104, line 1: As you explained to me earlier. You now change the logic, instead of confirming that we want to use account 'X' we are automatically logged in. Otherwise, we have to go through the profile completion screen

Yassyass181104 (approved):
This is way better like this since the greetings screen was causing some bugs overwritting the profile picture. This solves this bug and as I can see you improve the quality of the code and the UI. Very good job !

PR #397, opened 2024-12-15
Feature/auth/remove greetings screen

aminebenji04:
In this PR, you can see that I changed my mind a lot of times. I first started by removing the greetings screen but once that I have done this I realized tha we can access the app with an account created but with an incomplete profile. So my solution was to merge profile creation screen and sign up screen so that an account cannot be created without a complete profile.
### What I did ?
- Fixed a flicker issue where the "choose account" screen would briefly appear due to delays in database profile fetching.
Removed the "choose account" screen to improve app fluidity.

- Added a parameter to the loading screen and refactored its usage across all classes for consistency.
- Merged the sign-up screen and profile creation screen for a seamless onboarding experience.
- Removed the standalone profile creation screen as it is now integrated with the sign-up screen.
- Updated and refactored tests to align with the new implementation of the sign-up flow.
Added new tests to cover the merged screens and ensure robustness.
### How does it look now ?
- Enhanced user experience by eliminating redundant screens and ensuring a fluid onboarding flow.
- Improved code structure and test coverage for the authentication process.


app/src/androidTest/java/com/android/sample/ui/authentication/SignUpScreenTest.kt
Yassyass181104, line 1: As I can see you kind of refactored the tests such that it matches with the new logic used in here. Very good

app/src/main/java/com/android/sample/ui/components/LoadingScreen.kt
Yassyass181104, line 1: Modularizes the message and doesn’t put any if not specified which will not impact the rest of the code. Nice !

Yassyass181104 (commented):
You solved a big issue with this PR which is the problem that when opening the app we end up with the proposition of completing the profile even thought we already have done it. It created a bug of overriding the profile picture and other infos. Very nice correcting this !

Yassyass181104 (approved):

PR #399, opened 2024-12-16
Refactor/map screen UI

### LINKED ISSUE #400: Improve map screen based on the figma

elvan03:
In this PR,  I update the UI of the map screen. 

1) I have customised the map to look darker than the default. This matches better with the style of our app + makes pins(markers) more visible. The chosen colours for the map are available in the map_style.json file.

2) I have added a file (MapMarkerUtils) that allowed me to customise the pins(markers) on the map. Now, the markers have a colour that is defined by the category they belong to, for example, green if it's a Sport etc...

I have added a test file and have > 80% in coverage on new code.

closes #400


app/src/main/java/com/android/sample/ui/map/MapMarkerUtils.kt
Yassyass181104, line 1: This implementation of our own pin looks correct and very simple to use

app/src/main/java/com/android/sample/ui/map/Map.kt
Yassyass181104, line 1: You use the new pin implementation to have them colored depending on the category, very good.

Yassyass181104 (changes requested):
Overrall the implementation looks very good but the only problem I can point is that the tests you have created must be moved to the unit tests folder but they seem to be complete

elvan03:
Thanks for your review Yassine! The test file is already in the unit tests folder :))

Yassyass181104 (approved):

Yassyass181104:
> Thanks for your review Yassine! The test file is already in the unit tests folder :))

My bad, thanks for the clarification

PR #403, opened 2024-12-17
Fix strings displays as id instead of strings

Yassyass181104:
This pull request includes changes to the `ActivityForm` function in the `ActivityForm.kt` file to improve the way string resources are accessed. The changes ensure that string resources are fetched using the `context.getString` method instead of converting resource IDs to strings directly.

Improvements to string resource access:

* Changed validation for `title` to use `context.getString` for error messages.
* Changed validation for `description` to use `context.getString` for error messages.
* Changed validation for `price` to use `context.getString` for error messages.
* Changed validation for `placesMax` to use `context.getString` for error messages.
* Changed validation for `locationQuery` to use `context.getString` for error messages.


aminebenji04 (approved):
LGTM, this was a bit disturbing error, thanks for this

PR #408, opened 2024-12-17
Fix/auth/sign in google/no navigation profile edition

aminebenji04:
In this PR, I replaced the deletion of the ProfileCreation screen and took the old implementation of it. The reason of this was that I forgot that when creating an account with Google, it was navigating to the profile creation screen. Previously I replaced the profile creation screen by the sign up screen but this doesnt make sense when logging in with Google.

### DIFF
--- a/app/src/androidTest/java/com/android/sample/ui/profile/ProfileCreationTest.kt
+++ b/app/src/androidTest/java/com/android/sample/ui/profile/ProfileCreationTest.kt
@@ -38,7 +38,6 @@ class ProfileCreationTest {
   private lateinit var mockImageRepository: ImageRepositoryFirestore
 
   // Mock or create a fake ProfilesRepository
-  // private val mockProfilesRepository: ProfilesRepository = mock()
   @get:Rule val composeTestRule = createComposeRule()
 
   @Before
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/android/sample/ui/profile/ProfileCreation.kt
```diff
@@ -0,0 +1,271 @@
...
+              }
+
+          errorMessage?.let {
+            Text(
+                text = it,
+                color = Color.Red,
+                modifier = Modifier.padding(top = STANDARD_PADDING.dp).testTag("errorMessage"))
+          }
+        }
+  }
```
GhaliBerbich, line 270: LGTM, you just put back this file and its test file, nothing to add.

app/src/androidTest/java/com/android/sample/ui/profile/ProfileCreationTest.kt
```diff
@@ -0,0 +1,191 @@
...
+
+  private lateinit var navigationActions: NavigationActions
+  private lateinit var mockProfilesRepository: ProfilesRepository
+  private lateinit var profileViewModel: ProfileViewModel
+
+  private lateinit var mockImageViewModel: ImageViewModel
+  private lateinit var mockImageRepository: ImageRepositoryFirestore
+
+  // Mock or create a fake ProfilesRepository
+  // private val mockProfilesRepository: ProfilesRepository = mock()
```
GhaliBerbich, line 41: Please remove this unused line.
```suggestion
```

GhaliBerbich (changes requested):

GhaliBerbich (approved):

PR #409, opened 2024-12-17
fix(activity-details): add condition and remove edition button from t…

aminebenji04:
In this pull request, I fixed the bug of an activity being editable even when it is past.
The changes were simple and I onlty had to replace the button to edit by a text informing that the activity is past.


elvan03 (approved):
LGTM!

PR #411, opened 2024-12-17
Refactor: createActivity and editActivity for inconsistencies

### LINKED ISSUE #336: remove incoherent things in activity creation, e.g: number of person: 0... each field should make sense
Check that we cannot put impossible values:
- strings instead of numbers
- floats instead of ints
- Negative values
- What happens when we write more than the limit ?

elvan03:
In this PR, I fix the inconsistencies we had in edit and create activity. For example, we could give a number of participants of zero or put strings instead if numbers for max number of places and price fields.

I also solved an issue about "start time must be before end time" toast that was still displaying even though we changed "end time" variable by "duration" now. 

Now, we check entered value:
- is not negative (can be 0 for the price, not for maxPlaces)
- is a number only, no characters (can be double for price and is int for maxPlaces)
- is not empty

closes #336

### DIFF
--- a/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
+++ b/app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
@@ -332,14 +332,18 @@ fun CreateActivityScreen(
                               Toast.LENGTH_SHORT)
                           .show()
                       return@Button
-                    } else if (price.isNotBlank()  || price.toDoubleOrNull() == null || price.toDouble() < 0) {
+                    } else if (price.isNotBlank() ||
+                        price.toDoubleOrNull() == null ||
+                        price.toDouble() < 0) {
                       Toast.makeText(
                               context,
                               context.getString(R.string.invalid_price_format),
                               Toast.LENGTH_SHORT)
                           .show()
                       return@Button
-                    } else if ( placesMax.isNotBlank() || placesMax.toLongOrNull() == null ||  placesMax.toLong() <= 0) {
+                    } else if (placesMax.isNotBlank() ||
+                        placesMax.toLongOrNull() == null ||
+                        placesMax.toLong() <= 0) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
```diff
@@ -325,38 +325,35 @@ fun CreateActivityScreen(
...
-                          .show()
-                      return@Button
                     } else if (creator == "") {
                       Toast.makeText(
                               context,
                               context.getString(R.string.login_check_in_create),
                               Toast.LENGTH_SHORT)
                           .show()
                       return@Button
-                    } else if (!hourDateViewModel.isBeginGreaterThanEnd(startTime, duration)) {
```
aminebenji04, line 342: I don't really understand why you remove dthis condition.

aminebenji04 (changes requested):
I blocked merging because a part of the logic has been removed, I pointed it out into the comments. Let me know if you think this was useless

app/src/main/java/com/android/sample/ui/activity/CreateActivity.kt
elvan03: This condition had to be removed because we don't have an "end time" anymore, it was transformed into "duration". So basically, if I set start time at 7:00 and duration of activity is 2h, then this check would wrongly give the toast "start time must be before end time".

Let me know if it's more clear now, but it was necessary to remove this logic.

aminebenji04: OK now I understand

aminebenji04 (approved):
This is a good PR, I think that all the edge cases are blocked now, for the changes I requested, it was because I mastakenly understood how the duration property was working, LGTM !

PR #415, opened 2024-12-18
Improving line coverage for files below 80%

Yassyass181104:


GhaliBerbich (approved):
Thank you for your changes, it is now clearer with how you modularised code, and also good tests. I have nothing to add

PR #416, opened 2024-12-18
fix: remove possibility to like participants past activities

hmaryam84:
This small PR removes the possibility of liking and disliking the past activities of participants and also the possibility to see the likes and dislikes of participants.


aminebenji04 (approved):
Very good and concise PR, now we cannot change likes/dislikes of others, LGTM!

PR #423, opened 2024-12-18
Feature/choose default profile picture

GhaliBerbich:


app/src/main/java/com/android/sample/resources/C.kt
```diff
@@ -29,6 +29,7 @@ object C {
     const val ENROLLED_ACTIVITIES = 1
     const val PAST_ACTIVITIES = 2
 
+    const val CARD_IAMGES_SIZE = 120
```
hmaryam84, line 32: Very good for having added a constant in this file

hmaryam84: Note that there is another constant in this file with the same value (HALF_SCREEN_TEXT_FIELD_PADDING), but I think this is not problematic since they are used in completely different contexts

app/src/main/java/com/android/sample/ui/camera/Gallery.kt
hmaryam84, line 1: This logic is much better thant the old one. Well done!

app/src/main/java/com/android/sample/ui/dialogs/AddImageDialog.kt
```diff
@@ -100,12 +107,31 @@ fun AddImageDialog(onDismiss: () -> Unit, onGalleryClick: () -> Unit, onCameraCl
...
+                  Icon(Icons.Filled.AddToPhotos, contentDescription = "Default", tint = Color.Gray)
+                }
+          }
         }
       }
 }
 
 @Preview
 @Composable
 fun AddImageDialogPreview() {
```
hmaryam84, line 135: This preview was only for displaying the composable right? Maybe it is better to remove it

app/src/main/java/com/android/sample/ui/camera/Gallery.kt
hmaryam84: I reply to myself, after some researches, the @Preview are apparently ignored when the coverage is calculated, so it's fine to nkeep it!

hmaryam84 (approved):
Very good PR that implements an important feature and improves old UX issues. You wrote many interesting and coherent tests. Even though it seems to be hard to test (because it is in relation with captors,: camera, galery).  Also, you have a very good line coverage (96%). Even though there is some duplicated code, the overal work is well done! Looks Good To ME !

! Your PR has no detailed description which could really help, especially when the pr is heavy and changes many files.

hmaryam84:
Very good PR that implements an important feature and improves old UX issues. You wrote many interesting and coherent tests. Even though it seems to be hard to test (because it is in relation with captors,: camera, galery). Also, you have a very good line coverage (96%). Even though there is some duplicated code, the overal work is well done! Looks Good To ME !

! Your PR has no detailed description which could really help, especially when the pr is heavy and changes many files.

PR #424, opened 2024-12-18
Tests/improve line coverage

Yassyass181104:


app/src/test/java/com/android/sample/model/activities/ListActivitiesViewModelTest.kt
```diff
@@ -730,4 +739,102 @@ class ListActivitiesViewModelTest {
...
+            hourDateViewModel = hourDateViewModel,
+            startTime = "10:00",
+            duration = "1:00",
+            price = "10.0",
+            selectedLocation = activity.location,
+            selectedOptionCategory = Category.SPORT,
+            selectedOptionInterest = "Football")
+    assertTrue(result)
+  }
 }
```
GhaliBerbich, line 840: Thank you for adding these tests.

GhaliBerbich (approved):
Thank you for your work Yassine. I saw that in order to increase line coverage you removed all the log lines and added tests. If we were still developping the app, 
 and adding features I would have not accpeted to remove the logs, but now as we have tested our functionalities in real case scenarios, and we are preparing for the final submisson, I think that it is fair to remove these logs, as normally we should not be needing them as before, and also because we know our code and what may crash where.

PR #425, opened 2024-12-18
map(activity preview): display default picture in the preview

GhaliBerbich:
Quick Fix : 
- If the activity does not have images, it will display the category image corresponding.


app/src/main/java/com/android/sample/ui/map/Map.kt
```diff
@@ -283,19 +284,22 @@ fun MapScreen(
...
+    AsyncImage(
+        model = vignette,
+        contentDescription = "Activity image",
+        modifier =
+            Modifier.fillMaxWidth()
+                .height(LARGE_IMAGE_SIZE.dp)
+                .clip(RoundedCornerShape(TEXT_FONTSIZE.dp))
+                .testTag("activityImage"),
+        contentScale = ContentScale.Crop)
+    Spacer(modifier = Modifier.height(TEXT_FONTSIZE.dp))
```
aminebenji04, line 302: Very good, you use already defined constants and resolve the problem

aminebenji04 (approved):
Very good PR, fixes the issue linked to it

PR #431, opened 2024-12-18
Feature/caching profile picture

GhaliBerbich:
Overview
- This pull request introduces significant updates to how profile images are handled across the app. The updates simplify the codebase by removing outdated fields and enhance user experience by implementing image caching.

Changes
Refactoring Profile Management:
- The photo field is no longer used when updating user profiles across various screens (Edit Profile, Sign Up, Profile Creation). This change reflects a shift in how we handle user images, focusing on a more streamlined approach.
Caching Profile Pictures:
- Introduced caching for profile pictures to improve load times and user experience. When a profile picture is uploaded, it is now cached using SharedPreferences, which is also provided through dependency injection for easier access and testing.
- Updated ProfileImage composable to display the cached profile image while the new one is being fetched. This prevents the display gap where no image appears.
- Removed the functionality to delete a profile picture. Users who prefer not to use a personalized photo can now select from default options.


Yassyass181104 (approved):
You correctly redesign the aspect of a failure within the image viewmodel such that now in a case of failure the profile picture is fetched from the cache. Also you implement the saving of the PP in the cache. Very good, LGTM

PR #433, opened 2024-12-19
fix(gallery dismiss)

GhaliBerbich:
FIx this bug :
- When clicking on the button gallery in the add image dialog, if I dismissed, when I clicked again on the button gallery screen, nothing would appear, so it would not be possible to choose pictures. Now even after dismissing, I can choose pictures after.


Yassyass181104 (approved):
Very nice you fix an issue that was present from a long time and you correct it well. Good job !

PR #443, opened 2024-12-19
fix : images positions will not be swapped in activity details, and image displayed on activity card stays the same

GhaliBerbich:
If i go back then come again, the images may have a different order as their last because the fetching of the images returns the bitmaps in a "random" order, not predictable.
Edit : The solution used is to sort the Urls when fetched, then treat them sorted by TimeStamp order, when converting urls to bitmaps, so thht the returned list has the same order as the url list. The order is done according the timestamp of each picture, so that the ordering stays consistent. Now the list of images keeps the same order in activity details, even if fyou get out and go back the screen, also the first picture of an image which is displayed on the activity card stays the same thanks to this ordering.

The only problem, is that when an image is removed or added, all the other images are removed then put back, even though we did not touch them( this the logic implemented in upload activity images). So if I change the pictures, the order will change, but this new order will stay consistent everywhere until images are changed again. This cannot be resolved now because it would change the logic of uploading images for activities, so that could have been done if there was more time to work on it.

### DIFF
--- a/app/src/main/java/com/android/sample/model/image/ImageRepositoryFirestore.kt
+++ b/app/src/main/java/com/android/sample/model/image/ImageRepositoryFirestore.kt
@@ -224,7 +224,7 @@ constructor(private val firestore: FirebaseFirestore, private val storage: Fireb
         .addOnFailureListener { onFailure(it) }
   }
 
-  private fun sortUrlsByTimestamp(urls: List<String>): List<String> {
+  fun sortUrlsByTimestamp(urls: List<String>): List<String> {
     return urls.sortedBy { url ->
       // Extract the timestamp part from the URL
       url.substringAfterLast("image_").substringBefore('.').toLongOrNull() ?: 0L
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/android/sample/model/image/ImageRepositoryFirestore.kt
Yassyass181104, line 1: You implement a new way to sort the images such that it always keep the same ordering, very clever implementation

app/src/main/java/com/android/sample/ui/activity/EditActivity.kt
Yassyass181104, line 1: Here the same problem occurred but strangely you manage it correctly

app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
Yassyass181104, line 1: Cleaner and better implementation

Yassyass181104 (approved):
You solve a great issue by bringing a better implementation of the problem. Also you create tests to adjust the coverage very well done !

PR #449, opened 2024-12-19
Fix/default profile picture selector

GhaliBerbich:
- In this Pull Request, I reduced the size of the carousel where default images are displayed, so that the 2nd image can appear only in half suggesting that it is scrollable.
 - Also I reduced the size of the displayed images in carousel because they were too big.


Yassyass181104 (approved):
This resized version is way better and more intuitive for the user to know that the menu is scrollable, moreover the code is 100% covered, Great !!

PR #452, opened 2024-12-19
fix(profile): adjust logic to compute activities joined.

aminebenji04:
In this PR, I just updated the wrong logic that was done for computing the already joined activities:
```kotlin
it.creator != user.id || it.participants.map "{ it.id }.contains(user.id)
``` 
If you see here, the || condition is wrong and had to be replaced by a && condition so that an activity that contains our user id doesn't count as joined in all cases (if you create an activity, your uid will be in it and this logic will think that you have joined and not created the activity)


Yassyass181104 (approved):
Very nice, you indicate the separation between the actvities created and enrolled in such that we only count the ones that you haven't created. Good job !

PR #456, opened 2024-12-19
Fix/activity principal image overview

GhaliBerbich:
In this PR: 
- First Image of the activity is displayed on the activity card, if it exists, else a default image is displayed.
- The Image is displayed in such a way that favors horizontal images over vertical images.
- A choice was made to keep the implementation of lazyColumn to display the activityCards in the overview rather than column. While the latest permits to upload all images at once, when first arriving on the screen, if there are too many activities, this is not scalable and could create a crash, hence I decided to go with the more secure option, even though images might take a fraction of second before appearing when the activity was not on the screen before.

### Overview Screen
Capture d'écran 2024-12-20 001926

### LikedActivities Screen
Capture d'écran 2024-12-20 002040

### DIFF
--- a/app/src/main/java/com/android/sample/ui/listActivities/ViewListActivities.kt
+++ b/app/src/main/java/com/android/sample/ui/listActivities/ViewListActivities.kt
@@ -618,7 +618,3 @@ fun DarkGradient() {
                       startY = START_Y,
                       endY = END_Y)))
 }
-
-fun Bitmap.resizeToFixedSize(width: Int, height: Int): Bitmap {
-  return Bitmap.createScaledBitmap(this, width, height, true)
-}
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/android/sample/ui/listActivities/ViewListActivities.kt
Yassyass181104, line 1: This is where the magic happens, you have done a great job by putting the image of the activity in the overview which really makes the difference

app/src/main/java/com/android/sample/ui/listActivities/ViewListActivities.kt
```diff
@@ -598,3 +618,7 @@ fun DarkGradient() {
                       startY = START_Y,
                       endY = END_Y)))
 }
+
```
Yassyass181104, line 621: This method is useless in fact it should be deleted

Yassyass181104 (changes requested):
For me everything is good except the method that you have created that is useless

Yassyass181104 (approved):
You cleaned up the code and delete unused lines, LGTM

PR #464, opened 2024-12-20
fix/activityDetails/creator fetching

GhaliBerbich:
In this PR : 
- I fetch the creator informationswhen he is not in participant's list. Before, the only way to get the creator was to iterate on the participant's list, however it was introduced the possibility for the creator to leave the activity, or not be in it when in PRO. Now I have to call this method to getUserData when the creator is not in the participant's list.

- I Adjusted tests so that the clause where the creator is not in the list of participant never gets accessed in the tests because mocking this behavior is too complex and time worthy.

### DIFF
--- a/app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
+++ b/app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
@@ -164,9 +164,20 @@ fun ActivityDetailsScreen(
 
   val creatorID = activity?.creator ?: ""
   var creator by remember {
-    mutableStateOf(User(creatorID, "anonymous", "anonymous", listOf(), listOf(), "", listOf()))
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/sample/ui/activitydetails/ActivityDetailsScreen.kt
```diff
@@ -164,9 +164,20 @@ fun ActivityDetailsScreen(
...
+  if (creator.name == "null") {
+    profileViewModel.getUserData(
+        creatorID,
+        onResult = {
+          if (it != null) {
+            creator = it
+          }
+        })
+  }
+  Log.d("ActivityDetailsScreen", "ActivityDetails Creator: $creator")
```
aminebenji04, line 180: This was the missing logic, before this, it was displaying "Anonymous" when a creator is not enrolled in his own activity.

aminebenji04 (approved):
LGTM! It solves an important bug

PR #468, opened 2024-12-20
Providing 4th e2e test

mohamedtahaguelzim:
This PR features a new e2e test that signs a user in, makes them check that they have no liked activities, go like an activity, then go check if the activity has been liked.

It also features interesting flows, that have been put in comments. They should work, except for some minor test tags adjustments.

Their mocked backend should be already fully coded.


Yassyass181104 (approved):
Changes look good to me. The updates to the test repository module and mock profile repository handle the functionality as expected. Also, the adjustments in the UI tests and the addition of the new mock location repository are well implemented. The new end2end is very great and tests edge cases well done!

